PR #3, opened 2020-09-22
Add results page

tbarnahor:


karenav (approved):

src/main/webapp/style.css
```diff
@@ -0,0 +1,37 @@
+#content {
```
szwbar: I recommend giving explicit class names to the classes you're using, as you might have multiple lists or content classes. Same as you did with `.place-container`.

src/main/webapp/results.html
```diff
@@ -0,0 +1,44 @@
...
+<html>
+  <script src="script.js"></script>
+  <head>
+    <meta charset="UTF-8">
+    <title>capstone project</title>
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body onload="">
+    <div id="content">
+      <h1> Food Mood's recommendation for you:</h1>
```
szwbar: Redundant space?

src/main/webapp/results.html
```diff
@@ -0,0 +1,44 @@
...
+      <ol id="place">
+        <div class="place-container">
+        <li> Foo sushi </li> <br>
```
szwbar: Another indent?

src/main/webapp/results.html
```diff
@@ -0,0 +1,44 @@
...
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body onload="">
+    <div id="content">
+      <h1> Food Mood's recommendation for you:</h1>
+      <ol id="place">
+        <div class="place-container">
+        <li> Foo sushi </li> <br> 
+          <a href="https:www.fusushi"> <span title = "Fu sushi">Restaurant's website</span> </a><br> 
+          <small> <a href="https:www.googlemapsfusushi"> <span title = "Fu sushi on google maps">Open in google maps</span> </a> </small>
```
szwbar: I think you can use actual links :D

szwbar (approved):
Some comments. 
Seems OK! I don't have any access to the rendered page though :( Can you share a link that shows me how it looks?

src/main/webapp/results.html
tbarnahor: np.. I was lazy haha

PR #5, opened 2020-09-22
Add Place class and places sorter component

karenav:


src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
+// Copyright 2019 Google LLC
```
arielz-g: This class will be commonly used by several components, right? Consider having it in a /common directory so that it'll be approachable by all.

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.auto.value.AutoValue;
+import java.util.List;
+
+/**
+ * Represents a place that food can be ordered from.
+ */
+@AutoValue 
+public abstract class Place {
+  
+  /* The name of the place. */
```
arielz-g: /** for javadoc comment (same for the rest).

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+  public abstract String name();
+  
+  /* The url of the place’s website. */
+  public abstract String website();
```
arielz-g: nit: websiteUrl or just url.

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+  
+  /* A phone number that can be used to contact the place. */
+  public abstract int phone();
```
arielz-g: Is this validated as an int? How is this represented in Places API? Seems to me that there should be some sort of prefix (i.e. +972, etc.). If you support only Israeli phone numbers, notice that "0521234567" can't be saved as an int, and nor does "+972521234567". Consider String instead.

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+  
+  /* Coordinate of the physical place. */
```
arielz-g: Consider having this in a Credentials sub-class to reduce the amount of fields you've got here.

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+  public abstract Long longitude();
+  
+  /* Coordinate of the physical place. */
+  public abstract Long latitude();
+  
+  /* The types of cuisines the place offers. */
+  public abstract List<String> cuisines();
```
arielz-g: Do you want to allow free-form String here? How is this represented in the Places API? Perhaps an Enum will be more representative and can be widely used across your entire implementation (as you'd probably present a final amount of options in the form itself).

In addition, I presume that you don't want to allow duplication here, so perhaps a Set is a better collection to use.

src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,54 @@
...
+  
+  /* A short description of the place. */
+  public abstract String description();
+
+  /**
+   * Creates a new instance of the auto-value class.
```
arielz-g: auto-value is an implementation detail, not so much of a descriptive quality of the class. Think about what the class purpose is, for example: "Represents a place in the FoodMood app".

src/main/java/com/google/sps/data/PlacesSorter.java
```diff
@@ -0,0 +1,30 @@
...
+ * An interface for sorting Places according to predefined behaviour.
+ */ 
+interface PlacesSorter {
+  
+  /** 
+   * Sorts the given list according to predefined behaviour.
+   * @param unsortedPlaces: the list of places that we want to sort.
+   * @return the sorted list of places.
+   */
+  public List<Place> sortPlaces(List<Place> unsortedPlaces);
```
arielz-g: I'd just call this sort. Imagine the call would be placesSorter.sort(..), and not placesSorter.sortPlaces(..) - which is a given from the class name.

src/main/java/com/google/sps/data/PlacesSorter.java
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.List;
+ 
+/**
+ * An interface for sorting Places according to predefined behaviour.
+ */ 
+interface PlacesSorter {
```
arielz-g: General suggestion: implementing classes won't really have to keep state, right? That is, there isn't really much into the actual instance of the PlacesSorter (a good test for this is to ask yourself "can all class methods be static?").
In your case, the answer is yes, the sortPlaces method relies solely on it's input argument.

Consider the following alternative: make the Place class implement the Comparable interface (https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html). This obligates the class to implement the CompareTo method. You may then require an instance of a Comparator<Place> in the Place constructor, and apply it. The comparators can reside wherever you like, but for the random implementation it's just a matter of flipping an even coin between the two compared objects.
 
Sorting the list will then just be a matter of applying a regular java collection sort.

src/main/java/com/google/sps/data/RandomPlacesSorter.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Collections;
+
+/** 
+ * A PlacesSorter that sorts places randomly. 
+ */
+class RandomPlacesSorter implements PlacesSorter {
+
+    @Override
+    public List<Place> sortPlaces(List<Place> places) {
+      Collections.shuffle(places);
```
arielz-g: This changes your actual input (which is rarely done). My first advice would be making a copy and shuffle that one, leaving the origin untouched.
However, in Google we have Immutable types for that purpose exactly (to keep us from changing input sources), and I now see it's open source as well: https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html. Please change your APIs to use the Immutable types instead. This is a better practice overall.

src/main/java/com/google/sps/data/Place.java
karenav: That's a great point, I think we will think further about directories when we write the actual project with angular, because everything is going to change

src/main/java/com/google/sps/data/Place.java
karenav: Done :)

src/main/java/com/google/sps/data/Place.java
karenav: Done :)

src/main/java/com/google/sps/data/Place.java
karenav: that's a great point, thanks! changed :)

src/main/java/com/google/sps/data/Place.java
karenav: Done :)

src/main/java/com/google/sps/data/PlacesSorter.java
karenav: great comment, thanks!

PR #7, opened 2020-09-24
Place class and Places Sorter component 

karenav:


tbarnahor (approved):
Looks good to me. Just make sure all these files are in our food-mood project and not in the prototype.

prototype/src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,64 @@
...
+  /** The name of the place. */
+  public abstract String name();
+
+  /** The url of the place’s website. */
+  public abstract String websiteUrl();
+
+  /** A phone number that can be used to contact the place. */
+  public abstract String phone();
+
+  /** The rating of the place, represented by a number between 1-5. */
```
arielz-g: These two values have a predefined range as the documentation suggests. However, documenting it isn't enough (you should validate that this holds. Add a validation in the create(..) method for both, see https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html for helpful validation methods.

P.S: once this class contains non-trivial logic (like validating input), it should be accompanied with unit-tests.

prototype/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A utility class for Place objects.
+ */
+public final class Places {
+
+  /**
+   * A private constructor, so the utility class can't be instanciated.
+   */
+  private Places() { }
```
arielz-g: The common practice is to have a private constructor as the bottom-most method of the class.

prototype/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * A Place comparator that compares places randomly.
+   */
+  public Comparator<Place> randomComparator = new Comparator<Place>() {
```
arielz-g: This wasn't my exact intention when we've addressed the Places utility class.

This class should only contain static utility methods, specifically, have a generic sorting method:
public static ImmutableList<Places> sort(ImmutableList<Places> places, Comparator<Place> placeComparator) {..}

For the specifics, during call time, what I meant was to use:
 Places.sort(places, (place1, place2) -> \\ flip a coin);

If you'd like you can choose to implement Comparator classes instead of using a lambda \ have a method returning a random comparator as you've defined here \ have this comparator defined as a public static final.

prototype/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public int compare(final Place a, final Place b) {
+      Random rand = new Random();
+      return rand.nextBoolean() ? 1 : -1; //TODO: are these magic numbers,
```
arielz-g: Per your question, I now see a deeper problem for the specific random case: when using compartTo the outcome should be deterministic (i.e, if compareTo(a, b) == True, the result should remain the same for all consecutive calls). This won't work when randomizing a boolean parameter.

For this purpose alone, I suggest adding an additional method:
public static ImmutableList<Places> randomShuffle(ImmutableList<Places> places) {..}
Which you may utilize for this specific purpose (other comparisons would work with the other. LMK if you have any additional considered solutions.

karenav: great comment, thanks! I changed the function to work that way.

prototype/src/main/java/com/google/sps/data/Places.java
karenav: got it, thanks! This function no longer exists, and I'll keep in mind for the next ones.

prototype/src/main/java/com/google/sps/data/Places.java
karenav: great, moved it!

prototype/src/main/java/com/google/sps/data/Place.java
```diff
@@ -12,36 +12,63 @@
...
+  /**
+   * @return the place’s price level, represented by a number between 0-4.
+   */
   public abstract int priceLevel();
 
-  /** Coordinate of the physical place. */
+  /**
+   * @return Coordinate of the physical place.
+   */
   public abstract Long longitude();
```
arielz-g: Is it a long? A coordinate is usually a decimal number AFAIK.

prototype/src/main/java/com/google/sps/data/Place.java
```diff
@@ -54,11 +81,28 @@
...
+
+      final String ratingError = String.format(
+          "Rating should be between %s - %s", MIN_RATING, MAX_RATING
+          );
+      final String priceLevelError = String.format(
+          "Price level should be between %s - %s",
+          MIN_PRICE_LEVEL, MAX_PRICE_LEVEL
+          );
+
+      Preconditions.checkArgument(rating >= MIN_RATING && rating <= MAX_RATING,
```
arielz-g: use a static import.

prototype/src/main/java/com/google/sps/data/Place.java
```diff
@@ -54,11 +81,28 @@
...
   public static Place create(final String name, final String websiteUrl,
       final String phone, final int rating, final int priceLevel,
-      final Long longitude, final Long latitude) {
-    return new AutoValue_Place(name, websiteUrl, phone, rating, priceLevel,
-        longitude, latitude);
+      final Long longitude, final Long latitude)
+      throws IllegalArgumentException {
+
+      final String ratingError = String.format(
+          "Rating should be between %s - %s", MIN_RATING, MAX_RATING
```
arielz-g: These can be inline with the checkArgument statement. Notice you don't need String.format there - the method also excepts an error message template.

prototype/src/main/java/com/google/sps/data/Place.java
```diff
@@ -54,11 +81,28 @@
...
    * @return the new object instance.
+   * @throws IllegalArgumentException if an input isn't valid
+   *                                  (rating / price level)
    */
   public static Place create(final String name, final String websiteUrl,
       final String phone, final int rating, final int priceLevel,
-      final Long longitude, final Long latitude) {
-    return new AutoValue_Place(name, websiteUrl, phone, rating, priceLevel,
-        longitude, latitude);
+      final Long longitude, final Long latitude)
```
arielz-g: Why final (for all args in this file in general)? This isn't necessary \ common practice.

prototype/src/main/java/com/google/sps/data/Places.java
```diff
@@ -24,8 +27,13 @@
...
 
-  public static final ImmutableList<Place> randomSort(
-        ImmutableList<Place> places) {
+  /**
+   * Sorts the given list of Places randonly.
+   * @param places the list we want to sort.
+   * @return the randomly sorted list.
+   */
+  public static ImmutableList<Place> randomSort(
+      final ImmutableList<Place> places) {
```
arielz-g: No need for final (the class is already final).

prototype/src/main/java/com/google/sps/data/Places.java
```diff
@@ -24,8 +27,13 @@
...
     List<Place> mutablePlaces = new ArrayList<>(places);
```
arielz-g: You can use:
Lists.newArrayList(places);
Where Lists is from: https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Lists.html

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.sps.data.Place;
+
+@RunWith(JUnit4.class)
+public final class TestPlace {
+
+  /** A stub Place name. */
```
arielz-g: Aspire for tests to be standalone: don't use constants in tests unless these are common for all tests and don't make a difference in testing: i.e., each unit test is, well, a unit:) The tests should also be very readable - a reader should be able to understand what is the value that failed the test without looking around too much. 

To that end, I'd write the values INVALID_LOW_RATING \ VALID_RATING explicitly in the test. Example for invalid low rating:

Place validPlace = Place.create(NAME, WEBSITE, PHONE, 0, 3, LONGITUDE, LATITUDE);

My own opinion: I don't tend to use constants in tests unless they're common for all and their definition is very very complex. In this case the whole thing can be in place, no constants used, and then every test is completely standalone.

prototype/src/main/java/com/google/sps/data/Place.java
karenav: done :)

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
```diff
@@ -0,0 +1,102 @@
...
+  private static final int INVALID_HIGH_RATING = 10;
+  /** An invalid Place price level, that is too low. */
+  private static final int INVALID_LOW_PRICE_LEVEL = -1;
+  /** An invalid Place price level, that is too high. */
+  private static final int INVALID_HIGHT_PRICE_LEVEL = 5;
+
+  /**
+   * Test the Place creator: doesn't raise an error on a valid input.
+   */
+  @Test(expected = Test.None.class /* no exception expected */)
```
arielz-g: use AssertThat and assertThrows instead of (expected = ..).

See https://junit.org/junit4/javadoc/latest/org/junit/Assert.html - they both can be a static import. Example:
assertThrows(IllegalArgumentException.class, Place.create(...));

BTW I don't think you need a valid case here, it's a no-op (there's nothing to validate).

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
```diff
@@ -0,0 +1,102 @@
...
+  public void createPlace_validInput_noError() {
+    Place validPlace = Place.create(NAME, WEBSITE, PHONE, VALID_RATING,
+        VALID_PRICE_LEVEL, LONGITUDE, LATITUDE);
+  }
+
+  /**
```
arielz-g: No need for documentation for unit tests.

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
```diff
@@ -0,0 +1,102 @@
...
+    Place validPlace = Place.create(NAME, WEBSITE, PHONE, INVALID_LOW_RATING,
+        VALID_PRICE_LEVEL, LONGITUDE, LATITUDE);
+  }
+
+  /**
+   * Test the Place creator: raises an IllegalArgumentException if the given
+   * rating is too high.
+   */
+  @Test(expected = IllegalArgumentException.class)
+  public void createPlace_invalidHighRatingArg_raiseArgError() {
```
arielz-g: The convention is to use the method name as an infix, i.e.: create_invalidHighRating_throwsIllegalArgumentException.

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+package com.google.sps.test;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.sps.data.Place;
+
+@RunWith(JUnit4.class)
+public final class TestPlace {
```
arielz-g: PlaceTest (always use the class name and add "Test" suffix).

prototype/src/main/java/com/google/sps/data/Place.java
karenav: from what I've seen in the google maps API, they addressed it as a long

prototype/src/main/java/com/google/sps/data/Place.java
karenav: because of linter errors that demanded it to be final

prototype/src/main/java/com/google/sps/data/Places.java
karenav: same thing here, I got a linter error without it

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: I got linter errors without it

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: np, I can switch to that, but it would be great for me to understand - why is that a better practice?

prototype/src/main/java/com/google/sps/data/Places.java
karenav: cool, I didn't know this option! Is this a better practice?

prototype/src/main/java/com/google/sps/data/Place.java
karenav: that's a great idea, thanks! done

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: thank you for that input. I took all the specific constants inside the functions. 
left the ones that are used in all functions outside and renamed them, because I think that it's better to keep them unified. for example, if we change the rating system and now 3 becomes invalid, I think it makes more sense to change it in one place and not in all functions (there may even be more down the road). what do you think?

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: done

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: done

prototype/src/main/java/com/google/sps/data/Place.java
karenav: done

prototype/src/main/java/com/google/sps/data/Place.java
arielz-g: Shouldn't be that way, please remove.

prototype/src/main/java/com/google/sps/data/Places.java
arielz-g: Same

prototype/src/main/java/com/google/sps/data/Places.java
arielz-g: Generally prefer to use utility classes rather than writing your own logic. IDK if I'd call it a better practice, it's just better coding in general:)

prototype/src/main/java/com/google/sps/data/Place.java
karenav: but then the linter won't let me push the changes... It's blocked until all linter problems are fixed

prototype/src/main/test/java/com/google/sps/test/PlaceTest.java
```diff
@@ -0,0 +1,104 @@
...
+        VALID_RATING, VALID_PRICE_LEVEL, VALID_LONGITUDE, VALID_LATITUDE);
+  }
+
+  /**
+   * Test the Place creator: raises an IllegalArgumentException if the given
+   * rating is too low.
+   */
+  @Test(expected = IllegalArgumentException.class)
+  public void create_invalidLowRating_throwsIllegalArgumentException() {
+    final int INVALID_LOW_RATING = 0;
```
arielz-g: Use this inline (same for the rest). No need for a special variable (not do we use caps for internal method variables - this is reserved only for static final class fields).

prototype/src/main/test/java/com/google/sps/test/PlaceTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    Place validPlace = Place.create(VALID_NAME, VALID_WEBSITE, VALID_PHONE, 
+        INVALID_LOW_RATING, VALID_PRICE_LEVEL, VALID_LONGITUDE, VALID_LATITUDE);
+  }
+
+  /**
+   * Test the Place creator: raises an IllegalArgumentException if the given
+   * rating is too high.
+   */
+  @Test(expected = IllegalArgumentException.class)
```
arielz-g: Switch to use assertion methods as discussed (couldn't comment on prev conversation).

prototype/src/main/test/java/com/google/sps/test/PlaceTest.java
karenav: got it. The thing is I wanted to make it clear what this variable means. could I keep it with CamalCase?

prototype/src/main/java/com/google/sps/data/Place.java
karenav: done. hope the PR will be able to be pushed

prototype/src/main/java/com/google/sps/data/Places.java
karenav: done. hope the PR will be able to be pushed

prototype/src/main/test/java/com/google/sps/test/TestPlace.java
karenav: done. hope the PR will be able to be pushed

prototype/src/main/test/java/com/google/sps/test/PlaceTest.java
karenav: done :)

prototype/src/main/test/java/com/google/sps/test/PlaceTest.java
karenav: got it. without it the linter says I'm using magic nums. Also I wanted to make it clear what this variable means. So I kept it with CamelCase, what do you think?

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+/**
+ * The package that currently holds all the java files of the food-mood project.
+ */
+package com.google.sps.data;
+
+import com.google.auto.value.AutoValue;
+
+import static com.google.appengine.repackaged.
```
arielz-g: Is this because of the linter? (should be in one line)

arielz-g: Is this in two lines because of the linter? (should be in one line without a problem)

karenav: yeah, if it's in the same line it exceeds the 80 chars limit

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,104 @@
...
+   * @throws IllegalArgumentException if an input isn't valid
+   *                                  (rating / price level)
+   */
+  public static Place create(String name, String websiteUrl, String phone,
+      int rating, int priceLevel, Long longitude, Long latitude)
+      throws IllegalArgumentException {
+
+      checkArgument(rating >= MIN_RATING && rating <= MAX_RATING,
+          "Rating should be between %s-%s", MIN_RATING, MAX_RATING);
+
```
arielz-g: Please remove extra blank lines everywhere (commonly the code is written in a bulk). Both at method start and in between.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A utility class for Place objects.
+ */
+public final class Places {
+
+  /**
+   * Sorts the given list of Places randonly.
+   * @param places the list we want to sort.
+   * @return the randomly sorted list.
```
arielz-g: nit: a new list containing the original list's elements in random order.
Emphasize that the returned list is a fresh copy.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public static ImmutableList<Place> randomSort(ImmutableList<Place> places) {
+    List<Place> mutablePlaces = new ArrayList<>(places);
+    Collections.shuffle(mutablePlaces);
+    return ImmutableList.copyOf(mutablePlaces);
+  }
+
+  /**
```
arielz-g: no need for javadoc here (it's clear why it has a private constructor & private methods don't really require javadocs as they're not part of the public API).

If the linter obligates you to do so: add a rule that avoids requiring javadocs for private methods (and better yet, constructors).

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Collections;
+
+/**
+ * A utility class for Place objects.
+ */
+public final class Places {
```
arielz-g, line 28: Notice this class missing a unit test as well.

More of an FYI (and in case you noticed ignore my comment), but the reason that it's ok not to write unit tests for this method is that as this involves a random output (and you delegate the randomness to outer implementation), you can't know what the end result will be. However, if one so desires, you can validate that the output contains all elements as the input, or the sizes, etc. (IMO it's ok not to write one, but note that once you add another sort method to Places, a unit test is a must).

project/src/main/test/java/com/google/sps/test/PlaceTest.java
```diff
@@ -0,0 +1,81 @@
...
+  /** A valid Place price level. */
+  private static final int VALID_PRICE_LEVEL = 3;
+  /** A stub Place longitude. */
+  private static final long VALID_LONGITUDE = 35.35;
+  /** A stub Place latitude. */
+  private static final long VALID_LATITUDE = 30.30;
+
+  @Test
+  public void create_invalidLowRating_throwsIllegalArgumentException() {
+    final int invalidLowRating = 0;
```
arielz-g: Remove final, same for the rest.

arielz-g (approved):
Nice:) I'm approving this, as leftover comments are quite minimal. However, if you encounter a problem with one of these, feel free to consult with me.

project/src/main/test/java/com/google/sps/test/PlaceTest.java
karenav: done

project/src/main/java/com/google/sps/data/Places.java
karenav: done

project/src/main/java/com/google/sps/data/Places.java
karenav: great point, thanks. done

project/src/main/java/com/google/sps/data/Place.java
karenav: done

project/src/main/java/com/google/sps/data/Place.java
karenav: done

karenav: done

project/src/main/java/com/google/sps/data/Places.java
karenav: that's a great point. will add tests when we implement more sorting methods. thank you

project/src/main/test/java/com/google/sps/test/PlaceTest.java
karenav: when I remove it, I get an error that these are magic numbers..
why should the final be removed? from what I spoke to Bar it was good practice to declare final when the var won't change in the function

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -0,0 +1,95 @@
...
+   * @return the new object instance.
+   * @throws IllegalArgumentException if an input isn't valid (rating / price level)
+   */
+  public static Place create(String name, String websiteUrl, String phone,
+      int rating, int priceLevel, double longitude, double latitude)
+      throws IllegalArgumentException {
+      checkArgument(rating >= MIN_RATING && rating <= MAX_RATING, "Rating should be between %s-%s",
+          MIN_RATING, MAX_RATING);
+      checkArgument(priceLevel >= MIN_PRICE_LEVEL && priceLevel <= MAX_PRICE_LEVEL,
+      "Price level should be between %s-%s", MIN_PRICE_LEVEL, MAX_PRICE_LEVEL);
```
szwbar: 4 whitespaces

project/src/test/java/com/google/sps/test/PlacesTest.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.sps.data.Places;
+import static org.junit.Assert.assertTrue;
+
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
+
+@RunWith(JUnit4.class)
+public final class PlacesTest {
+
+  /** A valid Place object with name "name1". */
+  private static final Place VALID_PLACE_1 = Place.create(/*name*/ "name1", /*websiteURL*/
```
szwbar: Remove VALID ?

szwbar (approved):

project/src/main/java/com/google/sps/data/Place.java
karenav: done

project/src/test/java/com/google/sps/test/PlacesTest.java
karenav: done

PR #8, opened 2020-10-01
open project directory and move the initial angular project to draft

tbarnahor:
This push is for starting our project in a new "project file", organized in the same manner as the walkthroughs. The initial Angular project is saved in an unpushed draft folder.


karenav (approved):
Great, this is the base for our new work method! I'll add some changes when this is merged :)

arielz-g (approved):

PR #15, opened 2020-10-07
Places fetcher

tbarnahor:
Initial Version of PlacesFetcher, which still doesn't include:
Exception handling
The ability to search more than one type of cuisine
We may want to return some additional details that aren't returned in this version (Places status, Place cuisine types, etc.)


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.maps.model.PlaceType;
+import com.google.maps.model.PlacesSearchResponse;
+import com.google.maps.model.PriceLevel;
+import com.google.maps.GeoApiContext;
+import com.google.maps.TextSearchRequest;
+import com.google.maps.PlaceDetailsRequest;
+import com.google.maps.errors.ApiException;
+import com.google.maps.model.PlacesSearchResult;
+import com.google.maps.PlacesApi;
+import java.util.*;
```
karenav: from what I understand it's not recommended to import the whole package (.*), it's best to explicitly import the components that you need to use.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,149 @@
...
+     * Builds a query and requests it from Google Places API.
+     * 
+     * @return list of places that supply the query.
+     * @throws IOException
+     * @throws InterruptedException
+     * @throws ApiException
+     */
+    public List<Place> fetch() throws IOException, InterruptedException, ApiException { // TODO (talbarnahor): add exception handling and testing
+        PlacesSearchResult results[] = getPlacesSearchResults();
+        System.out.println(results.length);
```
karenav: notice the print

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,149 @@
...
+     * 
+     * @return A PlacesSearchResponse which contains the search results 
+     * @throws ApiException
+     * @throws InterruptedException
+     * @throws IOException
+     */
+    public PlacesSearchResult[] getPlacesSearchResults()
+            throws ApiException, InterruptedException, IOException {
+        TextSearchRequest query = 
+        PlacesApi.textSearchQuery(CONTEXT, cuisineType, location)
```
karenav: I think there needs to be an indentation of 4 spaces at the beginning of the line

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,149 @@
...
+     * @throws IOException
+     * @throws InterruptedException
+     * @throws ApiException
+     */
+    private List<Place> createPlacesList(PlacesSearchResult[] searchResultsArr)
+            throws ApiException, InterruptedException, IOException {
+        List<Place> places = new ArrayList<Place>();
+        for (PlacesSearchResult searchResult: searchResultsArr) {
+            PlaceDetails placeDetails = getPlaceDetails(searchResult.placeId);
+            Place place = Place.create(
```
karenav: Once the builder will be pushed, this should be altered to build a new Place (instead of create). You can see an example of how to use the builder in the Place test :)

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,116 @@
...
+  private static final PriceLevel priceLevel = PriceLevel.MODERATE; //used in PlaceDetails
+  private static final String placeId_1 = "ChIJN1t_tDeuEmsRUsoyG83frY4";
+  private static final String placeId_2 = "ChIJ02qnq0KuEmsRHUJF4zo1x4I";
+
+  /** Places to be used in tests. */
+  private static final Place PLACE_1 = Place.create("name1", urlString, phone, rating, priceLevelInt, location);
+  private static final Place PLACE_2 = Place.create("name2", urlString, phone, rating, priceLevelInt, location);
+
+  /** PlacesSearchResults to be used in tests. */
+  private static PlacesSearchResult SEARCH_RESULT_1 = createTestPlacesSearchResult(placeId_1);
```
karenav: same thing here, in the future will change to builder :)

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,116 @@
...
+  @Test
+  public void fetch_validSearchResults_returnsListOfPlaces() throws ApiException, InterruptedException, IOException {
+      PlacesFetcher spiedFetcher = spy(placesFetcher);
+      doReturn(PLACE_DETAILS_1).when(spiedFetcher).getPlaceDetails(placeId_1);
+      doReturn(PLACE_DETAILS_2).when(spiedFetcher).getPlaceDetails(placeId_2);
+      doReturn(SEARCH_RESULT_ARR).when(spiedFetcher).getPlacesSearchResults();
+      ImmutableList<Place> expectedOutput = ImmutableList.of(PLACE_1, PLACE_2);
+      assertEquals(expectedOutput, spiedFetcher.fetch());        
+    } 
+
```
karenav: I think this is a redundant space

tbarnahor: The linter thought so too :)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Great, I changed it to Build.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: right, I didn't notice it's missing. Fixed it

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Yeah thanks!

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: You are right :) Changed it

project/pom.xml
```diff
@@ -33,6 +33,16 @@
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
+    <dependency>
+        <groupId>com.google.maps</groupId>
+        <artifactId>google-maps-services</artifactId>
+        <version>0.15.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
```
szwbar, line 44: What's this dependency? Can you please document?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+import com.google.maps.PlaceDetailsRequest;
+import com.google.maps.errors.ApiException;
+import com.google.maps.model.PlacesSearchResult;
+import com.google.maps.PlacesApi;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
```
szwbar: Move import statement further up?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+    /** The type of places that will be searched. */
+    private static final PlaceType TYPE = PlaceType.RESTAURANT;
+
+    /** The search radius for place. */
+    private static final int SEARCH_RADIUS = 5000;
+    // TODO (M1): check at least 10 results, and if less extend radius
+
+    /** The entry point for a Google GEO API request. */
+    private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
+        .apiKey("AIza...") // TODO : save key in a file that can be accessed and pushed to github
```
szwbar: TODO(M0) ?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+    /** The cuisine types of the places that are fetched. */
+    private final String cuisineType;
+
+    /** The maximum price level as identified in Google Places of the fetched places. */
+    private final PriceLevel maxPriceLevel;
+
+    /** Specifies if the fetched places must be open at the time of fetching. */
+    private final boolean openNow;
+
+    /** The type of places that will be searched. */
```
szwbar: This comment is not very helpful. Prefer to explain here *why* `RESTAURANT` was chosen as the place to be searched.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
+
+public class PlacesFetcher {
+
+    /** Places will be fetched within a ceratain radius from this location. */
```
szwbar: 1. No need for JavaDoc on private fields. 
   If this is a linter error, we should remove the error. 
2. Why do you need `private` fields in this class? Do you intend to instantiate a single instance of this class for each request? Does this class need to preserve state along the request? 
    If not, why not just accept the request parameters in the `fetch` method, and then remove these private fields?  
   For the M0 stage, maybe it's better to hold the values statically (as you had before) as constants.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+    private static final int SEARCH_RADIUS = 5000;
+    // TODO (M1): check at least 10 results, and if less extend radius
+
+    /** The entry point for a Google GEO API request. */
+    private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
+        .apiKey("AIza...") // TODO : save key in a file that can be accessed and pushed to github
+        .build();
+
+    /**
+     * Fields are temporaraly hard coded for M0 version.
```
szwbar: nit: hardcoded (as a single word)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Queries Google Places API according to given params.
+     *
+     * @return A PlacesSearchResponse which contains the search results
+     * @throws ApiException
+     * @throws InterruptedException
+     * @throws IOException
+     */
+    public PlacesSearchResult[] getPlacesSearchResults()
```
szwbar: Why is this method public? 
It's a helper method for `fetch`, so it should be private (and does not require any Javadoc as a result).

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+            throws ApiException, InterruptedException, IOException {
+        TextSearchRequest query =
+            PlacesApi.textSearchQuery(CONTEXT, cuisineType, location)
+                .radius(SEARCH_RADIUS)
+                .maxPrice(maxPriceLevel)
+                .type(TYPE);
+        if (openNow) {
```
szwbar: From the `openNow` documentation in `TextSearchRequest`:
`@param openNow Whether to restrict this search to open places.`

In my understanding this means that if you provide "false" as input, the search is *not* restricted to (but may contain) places that are open now. In this case, I think you shouldn't have the condition here (`if (openNow) ...`).

Then, you can concatenate the call with `await()` which is more readable.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+            query.openNow(openNow);
+        }
+        PlacesSearchResponse results = query.await();
+        return results.results;
+    }
+
+    /**
```
szwbar: Javadoc is redundant on private methods.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+     * Builds a query and requests it from Google Places API.
+     *
+     * @return list of places that supply the query.
+     * @throws IOException
+     * @throws InterruptedException
+     * @throws ApiException
+     */
+    public List<Place> fetch() throws IOException, InterruptedException, ApiException {
+        PlacesSearchResult[] results = getPlacesSearchResults();
```
szwbar: Suggestion: make this a one-liner
`return createPlacesList(getPlacesSearchResults());`

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param searchResultsArr An array of maximum 20 PlacesSearchResults
+     * @return An immutable list of Places
+     * @throws IOException
+     * @throws InterruptedException
+     * @throws ApiException
+     */
+    private List<Place> createPlacesList(PlacesSearchResult[] searchResultsArr)
+            throws ApiException, InterruptedException, IOException {
+        List<Place> places = new ArrayList<Place>();
+        for (PlacesSearchResult searchResult: searchResultsArr) {
```
szwbar: add whitespace before colon (`:`)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+                Place.builder()
+                    .setName(placeDetails.name)
+                    .setWebsiteUrl(placeDetails.website.toString())
+                    .setPhone(placeDetails.formattedPhoneNumber)
+                    .setRating(placeDetails.rating)
+                    .setPriceLevel(Integer.parseInt(placeDetails.priceLevel.toString()))
+                    .setLocation(placeDetails.geometry.location)
+                    .build());
+        }
+        return ImmutableList.copyOf(places);
```
szwbar, line 112: Great!

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+import static org.mockito.Mockito.spy;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class PlacesFetcherTest {
+
+  /** url for Places used in tests. */
```
szwbar: nit: In comments we'd write in all caps: URL

Also, I don't believe these fields require documentation (and certainly not a JavaDoc).

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+  private static final String URL_STRING = "https://www.google.com/";
+
+  /** url for PlaceDetails used in tests. */
+  private static final URL URL = createTestURL(URL_STRING);
```
szwbar: Since you import URL as a type, I believe that having `URL` as a variable name is problematic.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+  /** Valid PlaceDetails for PLACE_1. */
+  private static final PlaceDetails PLACE_DETAILS_1 =
+    createTestPlaceDetails("name1", URL, PHONE, RATING, PRICELEVEL, LOCATION);
+
+  /** Valid PlaceDetails for PLACE_2. */
+  private static final PlaceDetails PLACE_DETAILS_2 =
+    createTestPlaceDetails("name2", URL, PHONE, RATING, PRICELEVEL, LOCATION);
+
+  /**
+   * creates URL from String for PlaceDetails used in tests.
```
szwbar: Nit: capital letter. Also in other comments. 
Also, there's no need for Javadoc on a private method.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+   *
+   * @param s The string representing the URL
+   * @return A URL
+   */
+  private static URL createTestURL(String s) {
+    try {
+      return new URL(s);
```
szwbar: What happens if you just have 

```java
private static final URL x = new URL(URL_STRING)
```
 Does this fail to compile? 
Otherwise I don't think you need this function at all.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+      return null;
+    }
+  }
+
+  /**
+   * creates PlaceDetails to be used in tests.
+   *
+   * @param name The name of the Place
+   * @param url The url of the Place's website
+   * @param phone The phonenumber of the Place
```
szwbar: nit: phone number

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+   * @param rating The Google rating of the Place
+   * @param priceLevel The priceLevel of the Place as identified by Google Places
+   * @param location The location of the Place
+   * @return The PlacesDetails according to the given params
+   */
+  private static PlaceDetails createTestPlaceDetails(String name, URL url, String phone,
```
szwbar: nit: add a newline after the first `(` for better formatting: 
```java
private static PlaceDetails createTestPlaceDetails(
    String name, URL url, String phone, float rating, PriceLevel priceLevel, LatLng location) {
```

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+    placeDetails.website = url;
+    placeDetails.formattedPhoneNumber = phone;
+    placeDetails.rating = rating;
+    placeDetails.priceLevel = priceLevel;
+    placeDetails.geometry = new Geometry();
+    placeDetails.geometry.location = location;
+    return placeDetails;
+  }
+
+  /** creates PlaceSearchResult to be used in tests.
```
szwbar: Newline after /**

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+   * @param placeId The place Id of the created PlaceSearchResult
+   * @return PlacesSearchResult with placeId as it's Id
+   */
+  private static PlacesSearchResult createTestPlacesSearchResult(String placeId) {
+    PlacesSearchResult searchResult = new PlacesSearchResult();
+    searchResult.placeId = placeId;
+    return searchResult;
+  }
+
+  /** A PlacesFetcher instance to be spied on. */
```
szwbar: I think that you more than spy on it - this is the class under test!

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -0,0 +1,157 @@
...
+    // TODO (talbarnahor): add exception handling and testing
+    /**
+     * Builds a query and requests it from Google Places API.
+     *
+     * @return list of places that supply the query.
+     * @throws IOException
+     * @throws InterruptedException
+     * @throws ApiException
+     */
+    public List<Place> fetch() throws IOException, InterruptedException, ApiException {
```
szwbar: You can return an `ImmutableList` as part of your API.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+import com.google.maps.model.PlacesSearchResult;
+import com.google.maps.model.PriceLevel;
+import com.google.sps.data.Place;
+import com.google.sps.data.PlacesFetcher;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
```
szwbar: Please decide whether you `import static` all mockito options (when, mock, etc.) or none.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+    return searchResult;
+  }
+
+  /** A PlacesFetcher instance to be spied on. */
+  private static PlacesFetcher placesFetcher = new PlacesFetcher();
+
+  @Test
+  public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
+    PlacesFetcher spiedFetcher = spy(placesFetcher);
+    doReturn(new PlacesSearchResult[0]).when(spiedFetcher).getPlacesSearchResults();
```
szwbar: Interesting solution! 
The problem that I see with the approach you took here, is that you won't be able to test the `getPlacesSearchResult` function. 
Instead, I think that you should have a single method that only calls `request.await()` and returns the value. Then you can mock out this method (and this method only), but still test your other method (the one in which you choose inputs). 
Same goes for the place details (you can even combine them). 
LMK if you want to talk about this f2f.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -0,0 +1,177 @@
...
+    ImmutableList<Place> expectedOutput = ImmutableList.of();
```
szwbar: Inline the expectation:
```java
assertEquals(ImmutableList.of(), ...);
```

szwbar (commented):
Great work! Mostly formatting comments and some suggestions. 

General question: why do you also push the `.class` files?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: The TODO is now done 👍 
I added the API key as an environment variable by adding 
            "env": {
                "API_KEY": "My API Key"
            }
to the launch.json file in .vscode folder

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Done

project/pom.xml
tbarnahor: It's a dependency on the Simple Logging Facade for Java (SLF4J) for the use of google-maps-services. Added documentation.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: 1. Yeah I added the javaDoc because of the linter errors. I will try to remove the error.
2. I added them as class fields to avoid passing them to the fetch() and from fetch() to getPlacesSearchResults(), but I understand now that it is better to pass them to each function and keep the ability to instantiate a single instance of this class.
I changed the values to be constants for M0 as suggested.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I changed it to public because otherwise I wasn't able to stub it in the tester. 
Anyway I will try to rearrange it like you suggested on the testing comment and maybe it will solve this visibility issue.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: sure

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: At first I understood it like you, but when testing I found that changing openNow to false is equivelent to seaching it as true, and only when I ommitted the querry.openNow I recieved results that include closed restaurants. 
You can see examples for this behavor [here on "NYC search examples"](https://docs.google.com/document/d/1D3O6nNO__P5a3gH8zC_TrIkc8ujWpRI7W4piUWYMWew/edit?usp=sharing), including an explanation about it from the official library documentation.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: Yes it fails to compile because of un handled exception. The best solution I found is to wrap it with try/catch in a function but I know it's weird.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: I ommitted them. Just left a short explanation about createTestURL because I think it's a wierd function so maybe it should be explained.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: true

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Removed it :)

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: I don't quite understand, can you please elaborate on that?
I should import options I am not using?

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: I understand and will change it. 
Just one dillema - do you think I should make an interface with a test implementation where those request.await() methods will be public so I can mock them, and production implementaion where they are private? Or it doesn't matter for them to be public?

project/pom.xml
szwbar: Ack, thanks. Interesting that they don't pull it in themselves...

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -25,52 +26,39 @@
...
-        .apiKey("AIza...") // TODO : save key in a file that can be accessed and pushed to github
+        .apiKey(System.getenv("API_KEY"))
         .build();
 
-    /**
-     * Fields are temporaraly hard coded for M0 version.
-     * In next versions those same fields will be the fields of a UserPrefrences
-     * instance passed to the PlacesFetcher constructor by the Servlet.
-     */
     public PlacesFetcher() {
```
szwbar: This constructor is now unneeded. Please remove it.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
szwbar: ?

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
szwbar: Ack, thanks.

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
szwbar: First, make the method package-private (aka default) and not public. 
Then, add an annotation of Guava's `@VisibleForTesting` on this method. This will help users of your class understand that this is not intended to be package-private, but must be because of testing. 
https://stackoverflow.com/questions/6913325/annotation-to-make-a-private-method-public-only-for-test-classes

project/src/main/java/com/google/sps/data/PlacesFetcher.java
szwbar: OK, let me know when you finish the test change.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -80,8 +68,7 @@ public PlacesFetcher() {
      * @throws ApiException
      */
     public List<Place> fetch() throws IOException, InterruptedException, ApiException {
```
szwbar: Continuing a conversation I cannot comment on anymore: 
why not accept the inputs as parameters in your library? and let the caller specify them? 
Otherwise you'll need another change to your library, although it's quite ready.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -95,35 +82,27 @@ public PlacesFetcher() {
...
         List<Place> places = new ArrayList<Place>();
-        for (PlacesSearchResult searchResult: searchResultsArr) {
-            PlaceDetails placeDetails = getPlaceDetails(searchResult.placeId);
+        for (PlacesSearchResult searchResult : searchResultsArr) {
+          PlaceDetails placeDetails = getPlaceDetails(searchResult.placeId);
             places.add(
                 Place.builder()
                     .setName(placeDetails.name)
-                    .setWebsiteUrl(placeDetails.website.toString())
+                    .setWebsiteUrl((placeDetails.website == null)
```
szwbar: nit: redundant parantheses.

szwbar (changes requested):
What about the extra .class files? Are they needed?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: It will be like that in M1 (as our [design](https://docs.google.com/document/d/1yPTb2Rh6MSTrDE9b9e1Q61ZCcH_cVNqBI8qs1VKNZEs/edit#) was planned. )
For now since in M0 we said it will all be hardcoded in the PlacesFetcher I don't see how inputs to fetch() are useful.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Sorry I missed it

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: Sound good, but this means the test has to be in the same package as PlaceFetcher and it's not how our packages are orginized. So you think I should put them in the same package or just use the annotation but with public visibility?

tbarnahor: I changed the functions like you suggested but I ran into a problem with testing the requests sent to the methods that call request.await() so IMO it's still not good enough. I pushed the best thing I got to and we can discuss this f2f and I explain the problem.

tbarnahor:
Regarding the .class files, they got pushed by accident. I removed them from this PR and also I added them to my local gitignore file so this won't happen again (tested it and they are indeed ignored from now on).

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -55,55 +55,55 @@
...
      *
+     * @param query A TextSearchRequest with all params to query on
      * @return A PlacesSearchResponse which contains the search results
      * @throws ApiException
      * @throws InterruptedException
      * @throws IOException
      */
-    public PlacesSearchResult[] getPlacesSearchResults()
+    @VisibleForTesting
+    public PlacesSearchResult[] getPlacesSearchResults(TextSearchRequest query)
```
szwbar: Since the test is in the same package as `PlacesFetcher`, a package-private modifier should be enough.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -112,25 +112,29 @@ public PlacesFetcher() {
...
+     * Queries Google Places API to recieve requested details about a certain place.
+     *
+     * @param request A PlaceDetailsRequest to query certain details on a certain place
+     * @return PlacesDetails containig requested details about the place
+     * @throws ApiException
+     * @throws InterruptedException
+     * @throws IOException
+     */
+    @VisibleForTesting
+    public PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
```
szwbar: Modifier

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
```diff
@@ -119,17 +121,19 @@ private static PlacesSearchResult createTestPlacesSearchResult(String placeId) {
...
   public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
     PlacesFetcher spiedFetcher = spy(placesFetcher);
-    doReturn(new PlacesSearchResult[0]).when(spiedFetcher).getPlacesSearchResults();
+    doReturn(new PlacesSearchResult[0])
+        .when(spiedFetcher).getPlacesSearchResults(any(TextSearchRequest.class));
     assertEquals(ImmutableList.of(), spiedFetcher.fetch());
   }
 
   @Test
   public void fetch_validSearchResults_returnsListOfPlaces()
```
szwbar: This comment is theoretical only:
Now that you have that spy in place (which is really close to an actual mock BTW!), you *could* test that the parameters that you provided are OK. 

You could add new tests that `getPlaceDetails` and  `getPlacesSearchResults` are called with the parameters you expect it to be called. This test your logic when building the request. 
For example, you could test that the given location matches the location you gave the function itself.

*However*, there are no getters on `TextSearchRequest` and `PlaceDetailsRequest`, sadly, so you will not be able to do that.

szwbar (approved):

project/src/test/java/com/google/sps/test/PlacesFetcherTest.java
tbarnahor: Yes, I really tried to do that for a while yesterday because it's bugging me that those methods are not really tested but I didn't find any way to get the info from the requests so I gave up for now.

karenav (approved):

PR #16, opened 2020-10-07
Servlet + javascript + index.html - connect the components and display in frontend

karenav:


project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int maxNumPlacesToRecommend = 3;
+    ImmutableList<Place> fetchedPlaces = PlacesFetcher.fetch();
+    ImmutableList<Place> sortedPlaces = Places.randomSort(fetchedPlaces);
+    int numPlacesToDisplay = Math.min(maxNumPlacesToRecommend, sortedPlaces.size());
+    response.setContentType("application/json");
+    for (int i = 0; i < numPlacesToDisplay; ++i) {
+      response.getWriter().write(new Gson().toJson(sortedPlaces.get(i)));
```
tbarnahor: Isn't there a way to convert a list to json without iterating over it's objects? And then send a single response containig all the info? I am not sure what is usualy done so whatever you think.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@WebServlet("/query")
+public final class QueryServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int maxNumPlacesToRecommend = 3;
+    ImmutableList<Place> fetchedPlaces = PlacesFetcher.fetch();
+    ImmutableList<Place> sortedPlaces = Places.randomSort(fetchedPlaces);
+    int numPlacesToDisplay = Math.min(maxNumPlacesToRecommend, sortedPlaces.size());
```
tbarnahor: Nice

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,43 @@
...
+ * A servlet that handles the user query. Currently accepts no input, and responds with a list of
+ * recommended places (in Json format).
+ */
+@WebServlet("/query")
+public final class QueryServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int maxNumPlacesToRecommend = 3;
+    ImmutableList<Place> fetchedPlaces = PlacesFetcher.fetch();
```
tbarnahor: At the end the fetch() function is not static (it was better that way for testing) so a PlacesFetcher instance should be instantiated. It will be enough to instantiate it once and use that instance for every request.

project/src/main/webapp/index.html
```diff
@@ -3,41 +3,62 @@
...
       </select>
       <br><br>
-      <input class="submit" type="submit" value="Decide for me!">
-    </form>
+      <label for="location" class="text-content">Location:</label>
+      <input class="button" type="text" id="location" name="location">
+      <br><br>
+      <input class="submit" type="submit" value="Decide for me!" onclick="fetchFromQuery()">
+    </div>
+    <div id="results" style="display: none">
```
tbarnahor, line 47: Good solution to the redirecting issue to put it all in one page. So if I understand correctly, regarding my task to make the results page all is left is to add the map?

project/src/main/webapp/script.js
```diff
@@ -0,0 +1,67 @@
...
+  const placeElement = document.createElement('div');
+  placeElement.class = 'place-container';
+
+  // Add name
+  const name = document.createElement('li');
+  name.innerText = place.name;
+  placeElement.appendChild(name);
+  placeElement.appendChild(document.createElement('br'));
+
+  // Add link to website
```
tbarnahor, line 43: I found cases where placeDestails.website is null on Google Places. For now I handled it by setting an empty string on the Place.websiteURL field. So in this case the user will see the 'Restaurant's website' title without it leading to a link.
We need to make sure that we either filter out places like that so only places that have all fields get here, or not display 'Restaurant's website' if it's url is "".
What do you think we should do?

project/src/main/webapp/index.html
karenav: yes exactly! :)

project/src/main/webapp/script.js
karenav: That's a great question. I think we'll have to make a decision about the first option, because it depends on other things, for example if we have a link to Google Maps that might be good enough. So for now I added a case that allows to display places without a website, and I think later on we'll decide about the filtering.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: great, thanks for the update! I added instantiation

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Good point. I didn't find a way to directly convert first n items of a list to json, but this inspired me to create a temp list with the desired top elements, and then convert it as a whole. 
Thanks!!

tbarnahor (approved):

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A servlet that handles the user query. Currently accepts no input, and responds with a list of
+ * recommended places (in Json format).
+ */
+@WebServlet("/query")
+public final class QueryServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int maxNumPlacesToRecommend = 3;
```
arielz-g: This should be a class private static final field.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,47 @@
...
+    ImmutableList<Place> fetchedPlaces = new PlacesFetcher.fetch();
```
arielz-g: The fetcher should also be a private final class field (initialize in constructor).

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,47 @@
...
+    //TODO(M1): add call to filterer
+    ImmutableList<Place> sortedPlaces = Places.randomSort(fetchedPlaces);
+    int numPlacesToDisplay = Math.min(maxNumPlacesToRecommend, sortedPlaces.size());
```
arielz-g: What happens if sortedPlaces.size() == 0? i.e., do you intend to display an empty results page (fine for now, just wondering)?

arielz-g: No need for this IMO, Stream.limit returns a stream consisting of the maximum elements, no longer than given size in length, of current stream (https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html).

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,47 @@
...
+    List<Place> placesToDisplay = sortedPlaces.stream()
```
arielz-g: Notice you can save up a lot of internal vars here by having this as a one liner:

placesToDisplay = Places.randomSort(placesFetcher.fetch()).stream().limit(MAX_RECOMMENDATIONS).collect(..)

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Done

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Done

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Absolutely right, thank you!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Done

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: That's a great question. It's something Tal and I put aside as a point to think about, so I think we should talk about it and implement it in M1. I will get back to you with an answer.

karenav:
I added two tests. The tests have a lot in common, actually - most of the functioning is the same except for numbers that can be injected from outside. So I was considering if I should use a helper function to avoid code repetition, or should it be repeated for readability, which is highly encouraged in test code.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import static org.mockito.Mockito.when;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.mockito.Mockito.mock;
```
arielz-g: import order should be:

static imports
// new line
// non-static imports

nit: maybe add a lint test for this?

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+public final class QueryServletTest {
+
+  private static final HttpServletRequest request = mock(HttpServletRequest.class);
+  private static final HttpServletResponse response = mock(HttpServletResponse.class);
+  private static final PlacesFetcher fetcher = mock(PlacesFetcher.class);
+  private static final int maxNumPlaces = QueryServlet.MAX_NUM_PLACES_TO_RECOMMEND;
+
+  @Test
+  public void getRequest_fetchedMoreThanMaxNumPlaces_respondWithMaxNumPlaces() throws Exception{
+    StringWriter sw = new StringWriter();
```
arielz-g: These are common to all tests. You can add a @Before method which executes prior to every test run.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
+import com.google.appengine.repackaged.com.google.gson.Gson;
+import com.google.appengine.repackaged.com.google.gson.JsonArray;
+import com.google.appengine.repackaged.com.google.gson.JsonObject;
+import com.google.maps.model.LatLng;
+
+@RunWith(JUnit4.class)
+public final class QueryServletTest {
+
+  private static final HttpServletRequest request = mock(HttpServletRequest.class);
```
arielz-g: Why is the request mocked?

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+  private static final HttpServletResponse response = mock(HttpServletResponse.class);
+  private static final PlacesFetcher fetcher = mock(PlacesFetcher.class);
+  private static final int maxNumPlaces = QueryServlet.MAX_NUM_PLACES_TO_RECOMMEND;
```
arielz-g: static final fields should be all caps, i.e. REQUEST, RESPONSE, MAX_NUM_PLACES, etc.

Worth having a lint check for this as well.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
arielz-g: I'd use explicit names (i.e. stringWriter, printWriter). sw and pw are hard to understand from context.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+    StringWriter sw = new StringWriter();
+    PrintWriter pw = new PrintWriter(sw);
+    int numOfFetchedPlaces = maxNumPlaces - 1;
+    ImmutableList<Place> placesListWithLessThanMaxNum = getImmutableListBySize(numOfFetchedPlaces);
+    QueryServlet servlet = new QueryServlet();
+    servlet.init(fetcher);
+    when(fetcher.fetch()).thenReturn(placesListWithLessThanMaxNum);
+    when(response.getWriter()).thenReturn(pw);
+
+    servlet.doGet(request, response);
```
arielz-g: These should be "batched" together (no new line separation). Notice that unit tests should be constructed as:
// Arrange

// Act

//Assert

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    JsonArray jsonPlaces = new Gson().fromJson(sw.getBuffer().toString(), JsonArray.class);
+
+    // Make sure we got the right number of elements
+    assertEquals(jsonPlaces.size(), numOfFetchedPlaces);
+  }
+
+  /**
```
arielz-g: nit: as this is a private method, you don't have to use javadoc style. A simple // comment explaining the method should do.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param numOfPlaces the number of elements we want to have on the list
+   * @return an immutable list that has the required number of Place elements. All elements are
+   * identical except for their name, which is serialized - '0', '1', '2', etc.
+   */
+  private static ImmutableList<Place> getImmutableListBySize(int numOfPlaces) {
+    List<Place> tempList = new ArrayList<Place>();
```
arielz-g: Use a builder here instead, i.e.:
ImmutableList.Builder<Place> places = ImmutableList.builder();
...
places.add(...)

return places.build();

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,107 @@
...
+    // Make sure we got the right number of elements
+    assertEquals(jsonPlaces.size(), numOfFetchedPlaces);
+  }
+
+  /**
+   * @param numOfPlaces the number of elements we want to have on the list
+   * @return an immutable list that has the required number of Place elements. All elements are
+   * identical except for their name, which is serialized - '0', '1', '2', etc.
+   */
+  private static ImmutableList<Place> getImmutableListBySize(int numOfPlaces) {
```
arielz-g: The word Immutable here is redundant (it's clear from the method's signature that it returns an immutable list, so that's more of an implementation detail, not contributing much to what it does - as the name should do).

In addition, a "get" prefix is used when you're, well, getting something that's already there (as opposed to create/build which is more fitting in this context).

I'd go for something like createPlacesListBySize.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: thanks for the clarification! I edited all the files that had imports. Checking about making a rule.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: sure, done :)

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Great, thanks

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: great, done

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Thank you for the clarification! done :)

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Great, done!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: I checked if there was a simple way to do it without mocking, and I didn't find one (using "new" doesn't work because it's an interface). I think that because we will soon want to mock it, that it's the right solution for now.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Done! Will check about the linter.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Done :)

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+  void init(PlacesFetcher inputFetcher) {
+    fetcher = inputFetcher;
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      //TODO(M1): add call to filterer
+      List<Place> placesToDisplay = Places.randomSort(fetcher.fetch())
```
arielz-g: nit: can be ImmutableList (there's a collector toImmutableList() as well).

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Test
+  public void getRequest_fetchedMoreThanMaxNumPlaces_respondMaxNumPlaces() throws Exception {
+    ImmutableList<Place> placesListWithMoreThanMaxNum =
+        createPlacesListBySize(QueryServlet.MAX_NUM_PLACES_TO_RECOMMEND + 1);
+    when(FETCHER.fetch()).thenReturn(placesListWithMoreThanMaxNum);
+
+    servlet.doGet(REQUEST, RESPONSE);
+
+    // Make sure we got the right number of elements
```
arielz-g: nit: can do without the comment (the test is self explanatory).

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -0,0 +1,104 @@
...
+          .setPriceLevel(3)
+          .setLocation(new LatLng(35.35, 30.30))
+          .build()
+      );
+    }
+    return places.build();
+  }
+
+  // Returns the number of json elements in the servlet's response
+  private int getResponseSize() {
```
arielz-g: Maybe getPlacesAmount? Response isn't that clear IMO

arielz-g (approved):
Nice!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Does it have an advantage if it's a local variable?

PR #18, opened 2020-10-08
change linter to accept no javadoc for private variables and methods

tbarnahor:


karenav (approved):

.github/linters/sun_checks.xml
```diff
@@ -89,10 +89,13 @@
       <property name="allowedAnnotations" value="Override, Test"/>
     </module>
     <module name="JavadocType"/>
-    <module name="JavadocVariable"/>
+    <module name="JavadocVariable">
+      <property name="excludeScope" value="private"/>
```
szwbar: It seems from the documentation that if you specify "public" scope (that should be the default), then the linter shouldn't complain about noon-puboic methods missing javadoc. 
https://checkstyle.sourceforge.io/config_javadoc.html#MissingJavadocMethod

Are you sure you need this? Can you share the problematic error that you're getting?

tbarnahor: These are the two kinds of errors I get:
image

That's why I added the "exclude scope" to both "JavadocVariable" and "MissingJavadocMethod".

szwbar: I see. 

JavadocVariable default scope is set to `private`. I think that instead of excluding `private` scope, you should set the scope to `public`. 

Line 58 is a constructor. It should be documented. So the `MissingJavadocMethod` change should be removed IMO.

tbarnahor: Got it, changed as suggested

szwbar (approved):

szwbar:
Thanks Tal!

PR #23, opened 2020-10-14
Multi cuisines fetching

tbarnahor:


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -37,7 +37,7 @@
      * will be fields of a UserPrefrences instance passed to fetch() by the Servlet.
      */
     private static final LatLng LOCATION = new LatLng(32.080576, 34.780641); // Rabin Square TLV;
-    private static final String CUISINES = "sushi"; // TODO(M0): change to set of types;
+    private static final List<String> CUISINES_LIST = new ArrayList<>(List.of("sushi", "burger"));
```
arielz-g: Use an ImmutableList here (always prefer to use immutable types).

arielz-g: I'd just call it CUISINES, the _LIST is redundant as it's given from the object type.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -66,7 +66,7 @@
      */
     public ImmutableList<Place> fetch() throws IOException, InterruptedException, ApiException {
         TextSearchRequest query =
-            PlacesApi.textSearchQuery(CONTEXT, CUISINES, LOCATION)
+            PlacesApi.textSearchQuery(CONTEXT, genCuisinesStr(CUISINES_LIST), LOCATION)
```
arielz-g: You're passing a static class variable here. For now, as the list isn't an outside parameter, i'd just change the method to be static, and use the static list for now.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,4 +137,15 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
             throws ApiException, InterruptedException, IOException {
         return request.await();
     }
+
+    private String genCuisinesStr(List<String> cuisineLst) {
```
arielz-g: This method name isn't that clear IMO. Always prefer to use explicit and understandable names both for methods and arguments (abbreviations aren't that common really).

Something like createCuisenesQuery sounds more fitting in this context.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,4 +137,15 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
...
+        String cuisineStr = "";
```
arielz-g: The whole method can be simplified by staticlly importing:
import static java.util.stream.Collectors.joining;

and then:
return CUISINES.stream().collect(joining("|"));

Always try to search for a utility to do simple things for you:)

karenav (commented):
Currently I don't have additional comments to what Ariel wrote :) 
The only thing I think is that we should use some sort of Enum for the cuisine types instead of strings. I'm not sure if it's in the scope of this PR or a different one, but I think it's something we should do.

tbarnahor:
> Currently I don't have additional comments to what Ariel wrote :)
> The only thing I think is that we should use some sort of Enum for the cuisine types instead of strings. I'm not sure if it's in the scope of this PR or a different one, but I think it's something we should do.

See my comment on the UserPrefrence PR regarding this. I have an open issue about this and it will be done as part of M1 PRs.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Makes sense

tbarnahor: Done

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I agree

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Thanks, I don't naturally think about using streams and other java utilities, I will keep it in mind fot next times.

karenav (approved):

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,4 +139,8 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
             throws ApiException, InterruptedException, IOException {
         return request.await();
     }
+
+    private static String createCuisenesQuery() {
```
szwbar: typo: cuisines

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,4 +139,8 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
...
+        return CUISINES.stream().collect(joining("|"));
```
szwbar: Test?

szwbar: By the way, I think this can be simplified by calling
`String.join("|", CUISINES`);`

I don't see the benefit of using streams here.

tbarnahor: Yes the join works, thanks.

Regarding the testing, isn't it enough that it is tested as part of the fetch testing? Since CUISINES is hardcoded in this version we only need it to work for this specific list of cuisines without any corner cases, and so this is tested in the current tests as part of fetch testing.

arielz-g (approved):

project/src/main/java/com/google/sps/data/PlacesFetcher.java
szwbar: The problem is that `getCuisinesQuery` is not tested now, and you'll probably not test it in the next PR (as it's an existing functionality). So the recommended way is to add the test directly when you add the functionality.

PR #24, opened 2020-10-14
add userPrefrences class

tbarnahor:


project/src/main/java/com/google/sps/data/UserPrefrences.java
```diff
@@ -0,0 +1,109 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.auto.value.AutoValue;
+import com.google.maps.model.LatLng;
+
```
karenav, line 20: There should be no space between these imports from what I understand (Ariel told me the favourable way to do this is by having a block of all static imports, then one new line, and then all the non-static imports).

project/src/main/java/com/google/sps/data/UserPrefrences.java
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * @return the coordinates of the location the user wants the food to be delivered to.
+  */
+  public abstract LatLng location();
+
+  /**
+  * @return the cuisine types that the user prefers.
+  */
+  public abstract List<String> cuisineTypes();
```
karenav: Did you consider making this an immutable list?

project/src/main/java/com/google/sps/data/UserPrefrences.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * @return a builder that enables to build a new Place object
+   */
+  public static Builder builder() {
+    return new AutoValue_UserPrefrences.Builder();
+  }
+
+
```
karenav: I think this is an extra space

project/src/main/java/com/google/sps/data/UserPrefrences.java
```diff
@@ -0,0 +1,109 @@
...
+     * @return the object that was built
+     */
+    abstract UserPrefrences autoBuild();
+
+    /**
+     * Concludes the building of a new UserPrefrences instance.
+     * @return the new instance.
+     * @throws IllegalArgumentException if an input isn't valid (rating / price level)
+     */
+    public UserPrefrences build() throws IllegalArgumentException {
```
karenav, line 99: A point to think about - we validate rating and price level. How about the other fields? Let's say a user didn't enter a location or any cuisines - is this valid? When we later use that information, do we consider that it might be empty? maybe we should add default values, like all cuisines for example...

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    /** The maximal valid rating value. */
+    private static final float MAX_RATING = 5.0f;
+    /** The minimal valid rating value. */
+    private static final float MIN_RATING = 1.0f;
+    /** The maximal valid price level value. */
+    private static final int MAX_PRICE_LEVEL = 4;
+    /** The minimal valid price level value. */
+    private static final int MIN_PRICE_LEVEL = 0;
+
+
```
karenav: I think this is an extra space

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@RunWith(JUnit4.class)
+public class UserPrefrencesTest {
+
+    private static final float RATING = 4;
+    private static final int PRICE_LEVEL = 2;
+    private static final LatLng LOCATION = new LatLng(32.08074, 34.78059);
+    private static final List<String> CUISINES = Arrays.asList("sushi", "burger");
```
karenav: I think we should think about using an Enum or something of that sort for the different cuisines, because now when we start to use them we would want to be able to address them in a unified way.

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+package com.google.sps.data;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import java.util.Arrays;
+import java.util.List;
+import com.google.maps.model.LatLng;
+import static org.junit.Assert.assertThrows;
```
karenav: Also here - from what I understand, static imports are supposed to be before the others, with a separating line

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
```diff
@@ -0,0 +1,92 @@
...
+            () -> assertEquals(RATING, prefrences.minRating()),
+            () -> assertEquals(PRICE_LEVEL, prefrences.maxPriceLevel()),
+            () -> assertEquals(LOCATION, prefrences.location()),
+            () -> assertEquals(CUISINES, prefrences.cuisineTypes())
+        );
+    }
+
+    /**
+     * @return a UserPrefrences builder that has valid values of all attributes
+     */
```
karenav: This is a private function, so it can be documented using "//" instead of javadoc (It's based on a comment I got from Ariel, and it makes a lot of sense for me because it separates the API public documentation from what we just want to clarify)

project/src/main/java/com/google/sps/data/UserPrefrences.java
tbarnahor: Right, I also saw it on the java style guide 👍

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
tbarnahor: Makes sense :)

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
tbarnahor: Absolutly. This regards the "add configuration file for multipe words textual search" task on our tasks list. I added an issue about it and will do it on a seperate PR on M1

project/src/main/java/com/google/sps/data/UserPrefrences.java
tbarnahor: Yes we should totaly think about that.

Regarding the locations we discussed yesteday and I understood we agreed that the prefrences form will always pass a valid location (the user can't submit the form until a valid location is entered, and it is checked by the Google Maps API used for getting the coordinates from the user's entered location). So the location validation already happened when reaching the UserPrefrences building. Do you think we should validate it again here?

Regarding cuisinses I will check the fetcher behavour when an empty textsearch string is passed. Anyway I think we should allow the cuisines list to be empty and this situation needs to be handeled at the fetcher, but it is a good point.

project/src/main/java/com/google/sps/data/UserPrefrences.java
tbarnahor: Right, even the linter missed it.

project/src/main/java/com/google/sps/data/UserPrefrences.java
tbarnahor: I didn't think about it, it does makes sense to make it immutable since it is for read-only uses. I changed it.

project/src/main/java/com/google/sps/data/UserPrefrences.java
karenav: When I mentioned the location I was thinking about the onboarding form for example, because as you said the user query location will be valid from the query page. I'm not sure we need to verify it, as long as we understand that the preferences don't have to be completely filled for every use. 
I thought a bit more about the cuisines and I think we should have a default that has all of them - because that's the actual default: if he didn't specify anything, he wants all of them. (BTW in the query form there is already logic that makes sure it isn't empty, but I still think we should do that, we can discuss why f2f if you want).

tbarnahor: Yes an empty cusines list means we don't restrict what cuisine types to search, so I think it is something that needs to be allowed and handled in the fetcher/ configiration file and not here or in the query form. Anyway IMO it doesn't regard this PR so we can continue this discussion offline.

project/src/main/java/com/google/sps/data/UserPrefrences.java
```diff
@@ -0,0 +1,107 @@
...
+  public abstract ImmutableList<String> cuisineTypes();
+
+  /**
+   * @return a builder that enables to build a new Place object
+   */
+  public static Builder builder() {
+    return new AutoValue_UserPrefrences.Builder();
+  }
+
+/**
```
szwbar: nit: missing space

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @param priceLevel the price level of a place, should be represented by a number between 0-4.
+     * @throws IllegalArgumentException
+     */
+    public static void validatePriceLevel(int priceLevel) throws IllegalArgumentException {
+        checkArgument(priceLevel >= MIN_PRICE_LEVEL && priceLevel <= MAX_PRICE_LEVEL,
+        "Price level should be between %s-%s", MIN_PRICE_LEVEL, MAX_PRICE_LEVEL);
+    }
+
+    private ValidationUtils() {
```
szwbar, line 51: Good. 
Usually the curly braces are in the same line
`private ValidateionUtils() {}`

project/src/test/java/com/google/sps/data/PlaceTest.java
```diff
@@ -52,24 +60,35 @@ public void create_invalidLowPriceLevel_throwsIllegalArgumentException() {
...
     });
   }
 
-  /**
-   * @return a Place builder that has valid values of all attributes
-   */
+  @Test
+  public void build_validInput_returnsValidPlace() {
+    Place place = getValidPlaceBuilder().build();
+      assertAll("place",
```
szwbar, line 74: Why do you need this `assertAll` syntax? 
Why not:
```java
assertEquals(NAME, place.name());
assertEquals(WEBSITE, place.websiteUrl());
...
```

project/src/test/java/com/google/sps/data/UserPrefrencesTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+      assertThrows(IllegalArgumentException.class, () -> {
+        getValidUserPrefrencesBuilder().setMaxPriceLevel(invalidHighPriceLevel).build();
+      });
+    }
+
+    @Test
+    public void build_validInput_returnsValidUserPrefrences() {
+        UserPrefrences prefrences = getValidUserPrefrencesBuilder().build();
+        assertAll("userPrefrences",
```
szwbar, line 74: Same here.

szwbar (approved):
LGTM! A few small comments.

project/src/test/java/com/google/sps/data/PlaceTest.java
tbarnahor: I read on some online articles that it is best to use assertAll when asserting a set of properties that belong together conceptionally. The advantage is that it always checks all of the assertions that are passed to it, no matter how many fail and then you get a detailed result of all that went wrong. Without it the first assertEquals that goes wrong will stop the test.

karenav (approved):

project/src/test/java/com/google/sps/data/PlaceTest.java
szwbar: Interesting! In Google that pattern it's not used, but I see the value in it. 
In C++ we have two different methods for testing: `EXPECT_THAT`, which does not stop the test if it fails; and `ASSERT_THAT`, which does stop the test. Then, we use `ASSERT` only if the condition we check is a precondition to other tests. For example:
```c
ASSERT_THAT(array, SizeIs(1)); // If this will continue execution, the next statement will fail since array[1] doesn't exist.
EXPECT_THAT(array[0], HasSomething());
```
There is something similar in Java but it's not widely used: https://truth.dev/api/0.42/com/google/common/truth/Expect.html

It's basically the same pattern, but it's used mainly for slow tests.

PR #27, opened 2020-10-20
PlacesFetcher gets UserPrefrences

tbarnahor:


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -61,18 +49,20 @@
     /**
      * Builds a query and requests it from Google Places API.
      *
+     * @param prefrences the UserPrefrences as specified by the user
      * @return an immutable list of places that supply the query
      * @throws FetcherException when an error occurs in querying the Places API
      *     for places or for places details
      */
-    public ImmutableList<Place> fetch() throws FetcherException {
+    public ImmutableList<Place> fetch(UserPrefrences prefrences) throws FetcherException {
```
karenav: I just noticed that "UserPrefrences" has a spelling error - should be "UserPreferences"

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -124,17 +135,45 @@ public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
...
+        .getPlacesSearchResults(any(TextSearchRequest.class));
+    assertEquals(
+      ImmutableList.of(PLACE_1, PLACE_2),
+      spiedFetcher.fetch(PREFRENCES_BUILDER.setCuisines(ImmutableList.of()).build()));
+  }
+
+  @Test
+  public void fetch_noOpenNowPreference_returnsListOfPlaces() throws Exception {
+    PlacesFetcher spiedFetcher = spy(placesFetcher);
+    doReturn(PLACE_DETAILS_1).
```
karenav: Nit: I think the dot should be at beginning on next line

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -124,17 +135,45 @@ public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
...
+        .getPlaceDetails(any(PlaceDetailsRequest.class));
+    doReturn(SEARCH_RESULT_ARR).when(spiedFetcher)
+        .getPlacesSearchResults(any(TextSearchRequest.class));
+    assertEquals(
+      ImmutableList.of(PLACE_1, PLACE_2),
+      spiedFetcher.fetch(PREFRENCES_BUILDER.setCuisines(ImmutableList.of()).build()));
+  }
+
+  @Test
+  public void fetch_noOpenNowPreference_returnsListOfPlaces() throws Exception {
```
karenav, line 172: I'm not sure I understand what this test checks. I understand that you don't require the results to be "open now", but all the mocked items are similar and are open by default (or maybe I'm missing something?), and the check is identical to the one in fetch_noPreferedCuisines_returnsListOfPlaces test. Wouldn't you want to check that when it's turned on you only get places that are open, and when it's not, you get all the places? (Here you have the second part, if I understand correctly).

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -124,17 +135,45 @@ public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
...
+    doReturn(SEARCH_RESULT_ARR).when(spiedFetcher)
+        .getPlacesSearchResults(any(TextSearchRequest.class));
+    assertEquals(
+      ImmutableList.of(PLACE_1, PLACE_2), spiedFetcher.fetch(PREFRENCES_BUILDER.build()));
+  }
+
+  @Test
+  public void fetch_noPreferedCuisines_returnsListOfPlaces() throws Exception {
+    PlacesFetcher spiedFetcher = spy(placesFetcher);
+    doReturn(PLACE_DETAILS_1).
```
karenav: Nit: I think the dot should be at beginning on next line

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Right, thanks

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Right, good catch!

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: This test checks that when openNow = false the program still sends the request and returns results. The resason for that is that there is no such thing to call the Places API with openNow = false, so when the places fetcher gets this it has to ommit the setting of the openNow param from the request. So actualy this test just checks that it is handled and the request is still sent. Unfortunatley I can't actually call the API in the unit test and check what is returned by it, so the only thing those tests checks is that the requests are set and sent (the logic around the fetching) and not the fetching itself.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -146,7 +136,7 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
         return request.await();
     }
 
-    private static String createCuisinesQuery() {
-        return String.join("|", CUISINES);
+    private static String createCuisenesQuery(ImmutableList<String> cuisines) {
```
szwbar: Typo somehow returned..

karenav (approved):

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -124,17 +135,43 @@ public void fetch_zeroSearchResults_returnsEmptyList() throws Exception {
...
       .getPlacesSearchResults(any(TextSearchRequest.class));
-    assertEquals(ImmutableList.of(), spiedFetcher.fetch());
+    assertEquals(ImmutableList.of(), spiedFetcher.fetch(PREFERENCES_BUILDER.build()));
   }
 
   @Test
   public void fetch_validSearchResults_returnsListOfPlaces() throws Exception {
     PlacesFetcher spiedFetcher = spy(placesFetcher);
-    doReturn(PLACE_DETAILS_1).doReturn(PLACE_DETAILS_2).when(spiedFetcher)
+    doReturn(PLACE_DETAILS_1).
```
szwbar: Formatting here should be:
Option 1: one line
```java
doReturn(foo).when(bar).getBaz();
```
Option 2:
```java
doReturn(foo)
  .doReturn(bar)
  .when(bar)
  .getBaz()
```

szwbar: (Same for other tests).

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
szwbar: Karen, your comment is valid, but due to API restrictions (we cannot mock the API itself), then some tests aren't complete, and we can't really do anything about it.

Tal, I think that in these kind of tests might benefit from building the TestSearchRequest itself: 

The same way you have a `PREFERENCES_BUILDER`, add a default `TextSearchRequest` instance with default values. 
Then instead of `any(TextSearchRequest.class)`, you need to expect this default `TextSearchRequest` with the parameter you expect. 
This way you'd actually test something in the test itself (test that when you insert `openNow = true` then the class actually generates a `TextSearchRequest` with `openNow = true`). 
Same goes for other tests of this kind.

szwbar (commented):
Generally LGTM, see comment about tests.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Great thanks, I wondered where the line breaks should be. Fixed to option two.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Yes because this was ready before the typo fix was merged and then when merging I stayed with this version. I will notice those things next times.

arielz-g (approved):

PR #29, opened 2020-10-20
Results page

tbarnahor:


project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -29,11 +29,11 @@
 
   /**
    * Sorts the given list of Places randomly.
-   * @param places the list we want to sort.
+   * @param immutableList the list we want to sort.
```
karenav: I don't understand the naming, the fact that it's an immutable list is reflected in the type, and I don't think the new name expresses the essence of the variable... What do you think?

project/src/main/webapp/script.js
```diff
@@ -68,5 +79,42 @@ function createPlaceElement(place) {
...
+
+
+/**
+ *  Creates a map and adds it to the page.
+ */
+function createMap() {
+  const USER_LOCATION = { lat: 32.080576, lng: 34.780641 }; //TODO(M1): change to user's location
+  var map = new google.maps.Map(document.getElementById('map-container'), {
+    center: USER_LOCATION,
+    zoom: 12,
```
karenav: Maybe we should keep the zoom level as a variable that adds a little bit of information, for example - whether it's close or far?

project/src/main/webapp/script.js
```diff
@@ -68,5 +79,42 @@ function createPlaceElement(place) {
...
+      map: map
+  });
+  if (place.websiteUrl) {
+    const infoWindow = new google.maps.InfoWindow({content: marker.description});
+    marker.addListener('click', () => {
+        infoWindow.open(map, marker);
+    });
+  }
+  // Zoom when clicking on marker
+  google.maps.event.addListener(marker,'click', function() {
```
karenav: Can also use "() => {}" insted of "function() {}". I personally like it better, but I don't know if one is preferred over the other.

karenav (approved):
Looks really good!

project/src/main/webapp/script.js
tbarnahor: right, I changed it like you suggested so it will be in the same format as other eventlisteners in the function.

project/src/main/webapp/script.js
tbarnahor: Good idea.

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: Oh I didn't notice the name change, maybe I accidently changed it but I don't remeber.. good that you noticed!
The only thing I meant to change is the ImmutableList import because it gave us errors that we used different imports for this in every file.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -38,7 +38,7 @@
      * the Servlet.
      */
     private static final LatLng LOCATION = new LatLng(32.080576, 34.780641); // Rabin Square TLV;
-    private static final ImmutableList<String> CUISINES = ImmutableList.of();
+    private static final ImmutableList<String> CUISINES = ImmutableList.of("sushi", "burger");
```
szwbar, line 40: How is this related to the PR?

project/src/main/webapp/script.js
```diff
@@ -12,19 +12,29 @@
...
   const placesDiv = document.getElementById('place');
-  fetch('/query').then(response => response.json()).then((places) => {
+  fetch('/query')
+  .then(response => response.json())
+  .then((places) => {
     places.forEach((singlePlace) => {
       placesDiv.appendChild(createPlaceElement(singlePlace));
-    });
-  });
+      addPlaceMarker(map, singlePlace)
```
szwbar, line 27: Something wrong with indentation here..

project/src/main/webapp/script.js
```diff
@@ -12,19 +12,29 @@
...
+      })
+    })
+  .then(() => {document.getElementById('waiting-message').style.display = 'none'})
+  .then(() => {document.getElementById('map-container').style.display = 'block'})
+  .then(() => {document.getElementById('feedback-box').style.display = 'block'});
```
szwbar: Why are these all in separate lambdas? 
I think that all of these should go under one method, something like - "displayMapAfterResults". 
I think that it can even include the entire logic after the 2nd "then".

szwbar (commented):
Please always provide a description to your pull requests.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: It isn't.. probably I changed it when I tested some things and didn't change back.

project/src/main/webapp/script.js
tbarnahor: Fixed it

project/src/main/webapp/script.js
tbarnahor: Yes it really should be one method, I didn't think about it. I changed it.

project/src/main/webapp/index.html
```diff
@@ -45,9 +46,10 @@ <h1> FoodMood - your food order helper!</h1>
       <input class="submit" type="submit" value="Decide for me!" onclick="fetchFromQuery()">
     </div>
     <div id="results" style="display: none">
-      <h2>Our recommendation:</h2>
+      <h2>Food Mood's recommendation:</h2>
+      <p id="waiting-message">Please wait as we are finding the best results for you</p>
       <ol id="place"> </ol>
-      <div id ="map-container"> </div>
+      <div id ="map-container" > </div>
```
arielz-g: nit: why the extra space?

arielz-g (approved):

project/src/main/webapp/index.html
tbarnahor: No reason, fixed it

PR #30, opened 2020-10-26
Filterer - add a filtering function + unit tests

karenav:


project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+   *        same place from the results. Places are considered the same if they have the same name.
+   * @return the filtered list.
+   */
+  public static ImmutableList<Place> filter(ImmutableList<Place> places, int minRating, Boolean
+      filterIfNoWebsite, Boolean filterBranchesOfSamePlace) {
+    //TODO(M1): also filter by place's status when the attribute is added
+    ImmutableList<Place> result = ImmutableList.copyOf(
+        places.stream()
+            .filter(place -> place.rating() >= minRating)
+            .filter(filterIfNoWebsite
```
tbarnahor: Here we wanted to filter out the place only if it has no website and no google link, so maybe even just add a TODO like you did with the status so we won't forget and leave it like this.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+   * filters the given list of places according to the given parameters.
+   * @param places the list we want to sort.
+   * @param minRating The minimal rating we want to limit the list by. A number between 1 and 5.
+   * @param filterIfNoWebsite Specifies whether we should filter if there is no available website
+   *        for the restaurant.
+   * @param filterBranchesOfSamePlace Specifies whether we should remove different branches of the
+   *        same place from the results. Places are considered the same if they have the same name.
+   * @return the filtered list.
+   */
+  public static ImmutableList<Place> filter(ImmutableList<Place> places, int minRating, Boolean
```
tbarnahor, line 58: I really like the way you implemented this, nice!

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
     return ImmutableList.copyOf(mutablePlaces);
   }
 
+  /**
+   * filters the given list of places according to the given parameters.
+   * @param places the list we want to sort.
```
tbarnahor: we are not sorting this list

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+      filterIfNoWebsite, Boolean filterBranchesOfSamePlace) {
+    //TODO(M1): also filter by place's status when the attribute is added
+    ImmutableList<Place> result = ImmutableList.copyOf(
+        places.stream()
+            .filter(place -> place.rating() >= minRating)
+            .filter(filterIfNoWebsite
+                ? place -> !Strings.isNullOrEmpty(place.websiteUrl())
+                : place -> true)
+            .filter(filterBranchesOfSamePlace
+                ? distinctByKey(p -> p.name())
```
tbarnahor: Sometimes "place" is used and sometimes "p" so I think to be consistent maybe keep using "place"

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+                : p -> true)
+            .collect(Collectors.toList())
+    );
+    return result;
+  }
+
+  //A utility function that allows to distinct beteen objects using a certain property
```
tbarnahor: between

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+  private static <T> Predicate<T> distinctByKey(Function<? super T, Object> keyExtractor) {
```
tbarnahor: In this implementation the place that apears first will be the one to stay in case of duplicates, right? 
I am not sure but didn't we want to keep only the place with the largest rating?

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +44,38 @@
...
+      Map<Object, Boolean> map = new ConcurrentHashMap<>();
+      return t -> map.putIfAbsent(keyExtractor.apply(t), Boolean.TRUE) == null;
```
tbarnahor: Why do we need to keep TRUE as value for every key in the map? Maybe it should just be a set and not a map if all values are always TRUE?

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -14,38 +14,23 @@
```
tbarnahor, line 14: The tests look good to me

project/src/main/java/com/google/sps/data/Places.java
karenav: Great point, I added the comment!

project/src/main/java/com/google/sps/data/Places.java
karenav: Thanks!

project/src/main/java/com/google/sps/data/Places.java
karenav: Absolutely right, fixed

project/src/main/java/com/google/sps/data/Places.java
karenav: Great point, changed

project/src/main/java/com/google/sps/data/Places.java
karenav: Thanks, changed

project/src/main/java/com/google/sps/data/Places.java
karenav: Good point, I added a sort so that the list will be sorted by rating

project/src/main/java/com/google/sps/data/Places.java
karenav: Actually I didn't write the function, I used it after I saw a lot of places that recommended using it. It's possible that your improvement is good, but I think I'll keep it that way because I know it works well. What do you think?

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: Great!

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,40 @@
     return ImmutableList.copyOf(mutablePlaces);
   }
 
+  /**
+   * filters the given list of places according to the given parameters.
```
arielz-g: Capital F.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,40 @@
...
+    //TODO(M1): also filter by place's status when the attribute is added
+    ImmutableList<Place> result = ImmutableList.copyOf(
+        places.stream()
+            .sorted(Comparator.comparing(Place::rating).reversed())
+            .filter(place -> place.rating() >= minRating)
+      //TODO(M1): take in account whether the place has a google maps link, when attribute is added
+            .filter(filterIfNoWebsite
+                ? place -> !Strings.isNullOrEmpty(place.websiteUrl())
+                : place -> true)
+            .filter(filterBranchesOfSamePlace
```
arielz-g: This one can be written as filter(filterBranchesOfSamePlace && distinctByKey(Place::name))
same for the other filter.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,40 @@
...
+                ? distinctByKey(place -> place.name())
+                : place -> true)
+            .collect(Collectors.toList())
```
arielz-g: Theres a toImmutableList collector: https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html.

project/src/main/java/com/google/sps/data/Places.java
arielz-g: On a different note, I'm not a fan of the distinctByKey usage. It's a hard read and I had to lookup an explanation myself to figure out what's going on here. For readability's sake, I propose you do the distinct phase using a set of some sort (similar to here: https://stackoverflow.com/questions/29670116/remove-duplicates-from-a-list-of-objects-based-on-property-in-java-8) or just separate the distinct operation from the one-liner.

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -14,38 +14,23 @@
...
-  private static final Place PLACE_1 =
-      Place.builder()
-          .setName("name1")
-          .setWebsiteUrl("website@google.com")
-          .setPhone("+97250-0000-000")
-          .setRating(4)
-          .setPriceLevel(3)
-          .setLocation(new LatLng(35.35, 30.30))
-          .build();
+  private static final Place PLACE_1 = getValidPlaceBuilder().setName("name1").build();
```
arielz-g: If you already have a utility method that retrieves a builder, use it directly in the test (without constants). 

Notice that test readers will look at PLACE_1 -> go to definition -> go to the method (and only then realize what's going on).

Notice that using constants can be helpful only when you're not hiding important test details from the reader.

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -54,4 +39,77 @@ public void randomSort_keepsAllItems() {
...
+        highRating /* min rating */,
+        false /* filter if no website */,
+        false /* filter branches of same place */
+    );
+
+    assertEquals(result, ImmutableList.of(highRatingPlace));
+  }
+
+  // Returns a Place builder that has valid values of all attributes.
+  private static Place.Builder getValidPlaceBuilder() {
```
arielz-g: I'd call this createValidPlaceBuilder(), notice that "get" is where you have a class instance and you want to get it, where here you're creating one.

arielz-g: Another comment regarding this method is that the only thing that matters here is the name (for test's sake). I'd make it a method argument so it's clear as sunshine that you're creating 3 valid places with the same name (for example).

You can name this createValidPlaceByName and return an instance instead of a builder.

If you want to change an argument during the test to distinguish the objects, you can do so by adding a toBuilder() method to Place class (which will return a Builder object of that instance), so for example in filter_fewBranchesOfSamePlace_filterOutBranches you can do:

ImmutableList.of(createValidPlaceByName("place 1").toBuilder().setRating(1).build(), createValidPlaceByName("place 2").toBuilder().setRating(2).build(), createValidPlaceByName("place 3").toBuilder().setRating(3).build());

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: I just don't understand what this TRUE value is used for but I am fine with leaving it like this if you prefer :)

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: Good idea. Can you push this sort? Or you want it in a seperate PR?

project/src/main/java/com/google/sps/data/Places.java
karenav: Got rid of the function, so I'm resolving :)

project/src/main/java/com/google/sps/data/Places.java
karenav: done

project/src/main/java/com/google/sps/data/Places.java
karenav: The second comment - done. 
The first comment - I'm not sure it makes sense to me. The function in the filter returns true when we want to keep the item, and false when we don't want to keep it. If I use this condition, then when we don't want to filter by the website (AKA when filterIfNoWebsite is false), the && condition will always be false and we'll filter everything out.

project/src/main/java/com/google/sps/data/Places.java
karenav: done

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: done

karenav: Great ideas, thanks. done

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: We do that even if it's used in a few different functions? We don't mind here about code repetition?

project/src/main/java/com/google/sps/data/Places.java
arielz-g: My bad, forgot to add an "!",see this condition: !(filterIfNoWebsite && Strings.isNullOrEmpty(place.websiteUrl()))
You can write down all cases and see it answers all scenarios correctly.

You can logically think about this as, what are the cases here (4 options for 2 parameters), then list down the desired outcome on each case. You'll then notice that the filter function should return false on only one case (filterIfNoWebsite == True and Strings.isNullOrEmpty(place.websiteUrl()) == True). The remaining cases should not filter this out.

Hope that helps:) Notice that what triggered my comment is that if you listed the filter condition as an if-else clause where one of the clauses just returns a boolean value - it can be simplified to one expression for sure.

project/src/test/java/com/google/sps/data/PlacesTest.java
arielz-g: You mean calling the method with the same argument in different tests? If that's it then yes, we don't mind it (if it makes sense).
Generally speaking in tests we prefer test readability over everything (and sometimes it does entail code repetition to a point, but that's where one might think what is irrelevant information to a test (and can be extracted to a method / constant) and what should remain there.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,31 @@
...
+   * @return the filtered list.
+   */
+  public static ImmutableList<Place> filter(ImmutableList<Place> places, int minRating, Boolean
+      filterIfNoWebsite, Boolean filterBranchesOfSamePlace) {
+    //TODO(M1): also filter by place's status when the attribute is added
+    ImmutableList<Place> result =
+        places.stream()
+            .sorted(Comparator.comparing(Place::rating).reversed())
+            .filter(place -> place.rating() >= minRating)
+      //TODO(M1): take in account whether the place has a google maps link, when attribute is added
```
arielz-g: Please have the comments indentation fit the code indentation.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,31 @@
...
+            .filter(filterIfNoWebsite
+                ? place -> !Strings.isNullOrEmpty(place.websiteUrl())
+                : place -> true)
+            .collect(collectingAndThen(toCollection(() -> new TreeSet<>(comparing(Place::name))),
```
arielz-g: Looks nicer:) Perhaps add a small comment above this that explains you're filtering out branches of the same place?
Also, where did filterBranchesOfSamePlace go?

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: No, that's not what I meant. I meant PLACE_1 and PLACE_2 are used in 3 different functions - so they will have the same code if it's not in an outer constant, but if that's preferred, then sure, done

project/src/main/java/com/google/sps/data/Places.java
karenav: ok got it, thanks! done

project/src/main/java/com/google/sps/data/Places.java
karenav: It's already pushed... I comment only after I do the change :)

project/src/main/java/com/google/sps/data/Places.java
karenav: haha you are absolutely right. Now I remembered why I didn't use this method in the first place (I ran into it when I was working on this) - because I didn't find an elegant way to add it with the condition. do you maybe have an idea? (aside for breaking the stream and starting a new one inside an 'if' statement).

project/src/main/java/com/google/sps/data/Places.java
karenav: done

project/src/main/java/com/google/sps/data/Places.java
arielz-g: Breaking the stream is fine IMO:)

tbarnahor (approved):

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -38,5 +45,32 @@
...
+      filterIfNoWebsite, Boolean filterBranchesOfSamePlace) {
+    //TODO(M1): also filter by place's status when the attribute is added
+    ImmutableList<Place> result =
+        places.stream()
+            .sorted(Comparator.comparing(Place::rating).reversed())
+            .filter(place -> place.rating() >= minRating)
+            //TODO(M1): also chack about having a google maps link, when attribute added
+            .filter(place -> !(filterIfNoWebsite && Strings.isNullOrEmpty(place.websiteUrl())))
+            .collect(ImmutableList.toImmutableList());
+        if (filterBranchesOfSamePlace) {
```
arielz-g: Please ident this properly.

arielz-g (approved):

project/src/main/java/com/google/sps/data/Places.java
karenav: done

PR #31, opened 2020-10-26
Take user location and choices from HTML 

karenav:


project/src/main/webapp/script.js
```diff
@@ -12,47 +12,106 @@
...
+  const quisines = document.getElementById('quisines-form').elements;
+  let result = '';
+  let i;
+  for (i = 0; i < quisines.length; i++) {
+    if (quisines[i].checked) {
+      result = result + quisines[i].value + ',';
+    }
+  }
+  if (result === '') {
+    throw new Error("Choose at least one quisine.");
```
tbarnahor: cuisine. It also need to be fixed in other places

project/src/main/webapp/script.js
```diff
@@ -12,47 +12,106 @@
...
+}
+
+function getUserRatingFromUi() {
+  return getCheckedValueByElementId('rating-form',
+      'Choose exactly one rating.');
+}
+
+function getUserPriceFromUi() {
+  return getCheckedValueByElementId('price-form',
+      'Choose exactly one price level.');
```
tbarnahor, line 69: Maybe we should explain the user that this is the maximum price level.

project/src/main/webapp/script.js
```diff
@@ -62,11 +121,131 @@ function createPlaceElement(place) {
...
+ * Displays the query form to the user and hides the results, so that the user can try to get new
+ * results.
  */
 function tryAgain() {
   document.getElementById('query-form').style.display = 'block';
   document.getElementById('results').style.display = 'none';
   document.getElementById('place').innerHTML = '';
 }
+
+function addSearchBoxToMap(map, input) {
```
tbarnahor: I don't quite understand what this function is doing, can you add a comment?
Is this the function that gets the location from the user and displays it on the map letting the user change it through the searchbox? If it is, why is it handeling a list of markers and places and not just one place the user enters? I mean in what situation this map handels more than one place and one marker shown on it?

project/src/main/webapp/script.js
```diff
@@ -62,11 +121,131 @@ function createPlaceElement(place) {
...
+  // Create the search box and link it to the UI element.
+  const searchBox = new window.google.maps.places.SearchBox(input);
```
tbarnahor: What is the input? A location? Maybe give it a more describing name

project/src/main/webapp/script.js
```diff
@@ -62,11 +121,131 @@ function createPlaceElement(place) {
...
+  map.controls[window.google.maps.ControlPosition.TOP_LEFT].push(input);
+  // Bias the SearchBox results towards current map's viewport.
+  map.addListener("bounds_changed", () => {
+    searchBox.setBounds(map.getBounds());
+  });
+  let markers = [];
+  // Listen for the event fired when the user selects a prediction and retrieve
```
tbarnahor: What does it mean "the user enters prediction" ?

project/src/main/webapp/script.js
```diff
@@ -62,11 +121,131 @@ function createPlaceElement(place) {
...
+      if (!place.geometry) {
+        return;
+      }
+      localStorage.setItem('userLocation',
+          JSON.stringify({lat: place.geometry.location.lat(), lng: place.geometry.location.lng()}));
+      // Create a marker for each place.
+      markers.push(
+        new window.google.maps.Marker({
+          map,
+          title: place.name,
```
tbarnahor: I think we will only use the marker on this map to mark the user's entered location, so why do we need place.name?

tbarnahor:
You did a really good job on getting the user choices from HTML and handeling errors regarding it, it is readable and covers all the cases.
Regarding the user location (especially the search box), I admit it was a bit difficult for me to understand, but I know it does the job so that's great. Maybe you can just see that it really includes just things that we need and it is more readable.

And in general I think for me it would have been easier to review this if it was a seperate PR handeling the user's location and map, but I guess it is individual.

project/src/main/webapp/script.js
karenav: Hahaha oh my god. Definitely

project/src/main/webapp/script.js
karenav: This is the marker of the location the user enters in the search box. The title in only so that the name of the place will appear if he hovers over the marker.

project/src/main/webapp/script.js
karenav: Good point, I made it clearer. Thanks!

project/src/main/webapp/script.js
karenav: This is not the function that gets the user's device location, it adds a search box to the map and defines everything that the search box needs to be able to do. I added a comment. 
The situation is for example when the user types something and presses "enter" without choosing a specific location, then the locations that fit that query the most are shown to him (as markers) and he can choose from them.

project/src/main/webapp/script.js
karenav: The query he enters to the search box. I changed the documentation

project/src/main/webapp/script.js
tbarnahor: But the location is his home adress, not some buisness or known place, so what is the name of these kind of places? I don't see where it's set.

project/src/main/webapp/script.js
tbarnahor: Oh ok, I didn't understand that has the functionality of showing several optional locations, now it makes more sense.

project/src/main/webapp/script.js
karenav: Again, this is not the user's location, it's the place that the google search returns (according to the user's input to the search box), and they have attributes like name etc :)

tbarnahor: Yes I understood that, but since the user's input to the search box is usually his home address and not a known place, I was wondering what name is in this attribute, and what kind of name we want it to have. We can deal with this when we work more on the UX, it's just something to pay attention to.

tbarnahor (approved):

project/src/main/webapp/script.js
karenav: It's just the name that comes from google maps, for example a string of his address

project/src/main/webapp/script.js
```diff
@@ -96,6 +154,128 @@ function tryAgain() {
   document.getElementById('place').innerHTML = '';
 }
 
+// Adds a search box to the map, and allows it to keep the user's updating location according to
+// his search box activity
+function addSearchBoxToMap(map, searchBoxElement) {
```
arielz-g, line 162: Can this be split to several sub-methods?

arielz-g (approved):

project/src/main/webapp/script.js
karenav: Did my best to make it more readable :)

PR #33, opened 2020-11-02
placeFetcher extend radius

### LINKED ISSUE #28: change radius in PlacesFetcher to grow depending in number of results

tbarnahor:
The purpose of this PR is to extend the fetching radius by 5km each time (up to 20 km) until at least 10 results are found.

### DIFF
--- a/project/src/main/java/com/google/sps/data/PlacesFetcher.java
+++ b/project/src/main/java/com/google/sps/data/PlacesFetcher.java
@@ -41,10 +41,10 @@ public class PlacesFetcher {
     private static final int INIT_SEARCH_RADIUS = 5000;
 
     /** The minimum number of results to be fetched. */
-    private static final int MIN_NUM_OF_RESULTS = 8;
+    private static final int MIN_NUM_OF_RESULTS = 10;
 
-    /** The maximum radius to search in is INIT_SEARCH_RADIUS * MAX_RADUIS_MULT. */
-    private static final int MAX_RADUIS_MULT = 4;
+    /** The maximum radius to search in is INIT_SEARCH_RADIUS * MAX_RADIUS_MULT. */
+    private static final int MAX_RADIUS_MULT = 4;
 
     /** The entry point for a Google GEO API request. */
     private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
@@ -60,28 +60,31 @@ public class PlacesFetcher {
      *     for places or for places details
      */
     public ImmutableList<Place> fetch(UserPrefrences prefrences) throws FetcherException {
-        TextSearchRequest query =
-            PlacesApi.textSearchQuery(
-                CONTEXT, createCuisenesQuery(prefrences.cuisines()), prefrences.location())
+        PlacesSearchResult[] PlacesSearchResult;
+        int radiusMult = 1;
+        do {
+            try {
+                PlacesSearchResult = getPlacesSearchResults(
+                    genTextSearchRequest(prefrences, INIT_SEARCH_RADIUS * radiusMult));
+            } catch (ApiException | InterruptedException | IOException e) {
+                throw new FetcherException("Couldn't fetch places from Places API", e);
+            }
+            radiusMult++;
+        } while (
+                PlacesSearchResult.length < MIN_NUM_OF_RESULTS && radiusMult <= MAX_RADIUS_MULT);
+        return createPlacesList(PlacesSearchResult);
+    }
+
+    private TextSearchRequest genTextSearchRequest(UserPrefrences prefrences, int radius) {
+        TextSearchRequest request = PlacesApi.textSearchQuery(
+            CONTEXT, createCuisenesQuery(prefrences.cuisines()), prefrences.location())
+                .radius(radius)
                 .maxPrice(PriceLevel.values()[prefrences.maxPriceLevel()])
                 .type(TYPE);
         if (prefrences.openNow()) {
-            query.openNow(prefrences.openNow());
+            request.openNow(prefrences.openNow());
         }
-        try {
-            PlacesSearchResult[] PlacesSearchResult;
-            int radiusMult = 1;
-            do {
-                query.radius(INIT_SEARCH_RADIUS * radiusMult);
-                PlacesSearchResult = getPlacesSearchResults(query);
-                radiusMult++;
-            }
-            while (
-                PlacesSearchResult.length < MIN_NUM_OF_RESULTS && radiusMult <= MAX_RADUIS_MULT);
-            return createPlacesList(PlacesSearchResult);
-        } catch (ApiException | InterruptedException | IOException e) {
-            throw new FetcherException("Couldn't fetch places from Places API", e);
-        }
+        return request;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -37,9 +37,14 @@
      */
     private static final PlaceType TYPE = PlaceType.RESTAURANT;
 
-    /** In this radius around "LOCATION" places will be searched. */
-    private static final int SEARCH_RADIUS = 5000;
-    // TODO(M1): check at least 10 results, and if less extend radius
+    /** In this radius around "LOCATION" places will be searched initially. */
+    private static final int INIT_SEARCH_RADIUS = 5000;
```
karenav: It's pretty obvious for us, but maybe it should be pointed out that it's in meters?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -55,20 +60,31 @@
...
-            query.openNow(preferences.openNow());
-        }
-        try {
-            return createPlacesList(getPlacesSearchResults(query));
-        } catch (ApiException | InterruptedException | IOException e) {
-            throw new FetcherException("Couldn't fetch places from Places API", e);
+            request.openNow(preferences.openNow());
         }
+        return request;
     }
```
karenav, line 88: The whole change looks really good!!
It took me a little while to understand what was going on, and I think it might have been more readable for me if MAX_RADIUS_MULT was called something else, like MAX_NUM_ITERATIONS_OF_RADIUS_CHANGE (or something else in that field).

tbarnahor: Right! changed it

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Right! changed it

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -37,9 +37,14 @@
      */
     private static final PlaceType TYPE = PlaceType.RESTAURANT;
 
-    /** In this radius around "LOCATION" places will be searched. */
-    private static final int SEARCH_RADIUS = 5000;
-    // TODO(M1): check at least 10 results, and if less extend radius
+    /** In this radius (in meters) around "LOCATION" places will be searched initially.*/
```
arielz-g: Use // commenting style for privates, /** is for javadoc purposes (same for other fields).
Also, perhaps name this as "INIT_SEARCH_RADIUS_MR" to suggest the unit.

nit for phrasing: The initial search radius in meters.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -55,20 +60,32 @@
      *     for places or for places details
      */
     public ImmutableList<Place> fetch(UserPreferences preferences) throws FetcherException {
-        TextSearchRequest query =
-            PlacesApi.textSearchQuery(
-                CONTEXT, createCuisinesQuery(preferences.cuisines()), preferences.location())
-                .radius(SEARCH_RADIUS)
+        PlacesSearchResult[] placesSearchResult;
+        int radiusMult = 1;
```
arielz-g: attemptsCounter?

arielz-g: Initialize this with 0 and change below to < instead of <=

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -209,4 +209,18 @@ public void fetch_PlaceDetailsQueryFails_throwsFetcherException() throws Excepti
     assertTrue(thrown.getCause() instanceof IOException);
     assertTrue(thrown.getMessage().contains("place details"));
   }
+
+  @Test
+  public void fetch_resultsOnlyAfterRadiusExtension_returnsListOfPlaces() throws Exception {
```
arielz-g, line 220: Add a test that retrieves nothing after max number of extensions was reached to handle the edge case.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Changed it.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I wanted it to be the factor the initial radius is multiplied by but indeed it is not clear, so I changed it like you suggested because one of his jobs it to also count attempts.

tbarnahor: I initialized it to 1 because it multiplies the initial search radius (which if first multiplied by 1). I changed it to be initiated to 0 and updated in the begining of the attempt so I hope now it's clearer.

And you are right about < , changed it.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: The fetch_zeroSearchResults_returnsEmptyList test checks this corner case. In this test the getPlacesSearchResults always returns nothing so this test ends only when max number of extensions were reached and recieves nothing.

karenav (approved):

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -31,17 +31,21 @@
...
+    // rather as "RESTAURANT" this is the most suitable type to search for.
     private static final PlaceType TYPE = PlaceType.RESTAURANT;
 
-    /** In this radius around "LOCATION" places will be searched. */
-    private static final int SEARCH_RADIUS = 5000;
-    // TODO(M1): check at least 10 results, and if less extend radius
+    // The initial search radius in meters.
+    private static final int INIT_SEARCH_RADIUS_M = 5000;
+
+    // The minimum number of results to be fetched.
```
arielz-g: nit: minimal

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -31,17 +31,21 @@
...
+    private static final int MIN_NUM_OF_RESULTS = 10;
+
+    // The maximum number of times the search radius will be extended.
```
arielz-g: nit: maximal

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
arielz-g: I see, in that case please add a verification in the test that the mock was called MAX_NUM_OF_RADIUS_EXTENSIONS. You can use mockito's verify method, something like:
Mockito.verify(spiedFetcher, times(MAX_NUM_OF_RADIUS_EXTENSIONS)).getPlacesSearchResults()

see https://javapointers.com/java/unit-test/use-verify-in-mockito/ for more info.

arielz-g (approved):
Approving with minor fixes, LMK if something's unclear:)

PR #34, opened 2020-11-02
Cuisines config file

### LINKED ISSUE #25: Add configuration file for cuisine types text search

tbarnahor:
Adds a configration file for mapping the cuisine name to other search words to be used for Places fetching.


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -39,20 +46,22 @@
...
     private static final int SEARCH_RADIUS = 5000;
-    // TODO(M1): check at least 10 results, and if less extend radius
 
     /** The entry point for a Google GEO API request. */
     private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
         .apiKey(System.getenv("API_KEY"))
         .build();
 
+    /** A mapping between cuisines and text search words. */
+    private static Map<String, String[]> cuisineToSearchWords = getCuisinesMap();
```
karenav: why do we keep this outside if it's only used in one place?

karenav (approved):
looks really good!

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I wanted to read the json and create the map once and reuse it for all queries (instead of calling the getCuisinesMap() on every query). Do you think it is better to do this differently? 
Anyway I changed it to have constant name.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -39,20 +46,22 @@
...
+    private static final Map<String, String[]> CUISINE_TO_SEARCH_WORDS_MAP = getCuisinesMap();
+
     /**
      * Builds a query and requests it from Google Places API.
      *
      * @param preferences the UserPreferences as specified by the user
      * @return an immutable list of places that supply the query
-     * @throws FetcherException when an error occurs in querying the Places API
-     *     for places or for places details
+     * @throws FetcherException when an error occurs in querying the Places API for
```
arielz-g: Why is this in here?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -95,8 +104,7 @@
             try {
                 placeDetails = getPlaceDetails(detailsRequest);
             } catch (ApiException | InterruptedException | IOException e) {
-                throw new FetcherException(
-                    "Couldn't get place details from Places API", e);
+                throw new FetcherException("Couldn't get place details from Places API", e);
```
arielz-g: Same as above.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -39,20 +46,22 @@
...
     private static final int SEARCH_RADIUS = 5000;
-    // TODO(M1): check at least 10 results, and if less extend radius
 
     /** The entry point for a Google GEO API request. */
     private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
         .apiKey(System.getenv("API_KEY"))
         .build();
 
+    /** A mapping between cuisines and text search words. */
+    private static final Map<String, String[]> CUISINE_TO_SEARCH_WORDS_MAP = getCuisinesMap();
```
arielz-g: _MAP is redundant, "CUISINE_TO_SEARCH_WORDS" is descriptive enough along with the type (as it's a map).

arielz-g: Does it have to be an array? Prefer ImmutableList.

arielz-g: Use ImmutableHashMap.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,6 +145,18 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
...
 
     private static String createCuisinesQuery(ImmutableList<String> cuisines) {
-        return String.join("|", cuisines);
+        return cuisines.stream()
+            .map(cuisine -> String.join("|", CUISINE_TO_SEARCH_WORDS_MAP.get(cuisine)))
+            .collect(Collectors.joining("|"));
+    }
+
+    private static Map<String, String[]> getCuisinesMap() {
+        System.out.println(PlacesFetcher.class.getClassLoader());
```
arielz-g: Remove?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -137,6 +145,18 @@ PlaceDetails getPlaceDetails(PlaceDetailsRequest request)
...
+        Type mapType = new TypeToken<Map<String, String[]>>() {
+        }.getType();
+        return new Gson().fromJson(new JsonReader(
+            new InputStreamReader(
+                PlacesFetcher.class.getResourceAsStream("cuisinesSearchWords.json"))),
```
arielz-g: Keep this in a named static final.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -47,7 +47,7 @@
   private static final float RATING = 4;
   private static final int PRICELEVEL_INT = 2; // used for Places and UserPreferences
   private static final PriceLevel PRICELEVEL = PriceLevel.MODERATE; // used for PlaceDetails
-  private static final ImmutableList<String> CUISINES = ImmutableList.of("sushi", "burger");
+  private static final ImmutableList<String> CUISINES = ImmutableList.of("sushi", "hamburger");
```
arielz-g, line 52: Missing a test for validating the configuration's use. Something like making sure that the string that was used in the query contained what it is that you want.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
karenav: Yeah, if you'll use it multiple times that'll make sense.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Haha right

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Same as above.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Do you mean why is this change is in this PR? Probably at some point when working on this PR it jumped to be on one line (the VScode does it for me somethimes) and when I broke the line again I broke it in a different place so it apears here as a change.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: The getCuisineMap() is only called once to generate the  private static final CUISINE_TO_SEARCH_WORDS map. So I think it will not spare calls for new InputStreamReader() to keep it as a static field.
Or did you mean to give it a name inside this function for better readability?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I changed it to ArrayList. I didn't find a straight forward way to read the json values into an immutableList. I can keep struggling with this, but I thought keeping the whole map as an immutableMap with ArrayList as values is good enough. What do you think?

tbarnahor: I changed the Map to immutableMap as i didn't find an ImmutableHashMap type.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: I added tests. It made me change the visibility of the createCuisinesQuery() function so let me know what you think about it.
Also adding the tests reminded me that I didn't handle invalid cuisines input so I added an exception throw for that case.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
arielz-g: Still has _MAP suffix on my end.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
arielz-g: I meant having the string "cuisinesSearchWords.json" as a static final.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
arielz-g: Can you revert to how it was?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
arielz-g: Don't see the change on my end. Regarding your question, getCuisinesMap is the one returning the value, it can be intermediately be read as an array (or something else) and then converted to an ImmutableList by a simple conversion (ImmutableList.copyOf(/* array */).

arielz-g: Changes aren't visible..

arielz-g (commented):
Hey, most changes you've made aren't visible on my end, so can't really comment on most questions. Please push and LMK when this is good to review.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Yes, I reverted

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Wierd, for me it apears here on github without the _MAP suffix.
Maybe not refreshed?

tbarnahor: Yes, but then I have to iterate over the map that getCuisinesMap returns and do this conversion for every entry on it.
I suppose there is a way to do this while reading the json instead of building the whole map and then going back to modify each entry, but I didn't find something that worked.
Is this extra map iteration and lists copying worth it? Isn't it enogh for the map to be immutable?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Ok, changed it

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -209,4 +209,26 @@ public void fetch_PlaceDetailsQueryFails_throwsFetcherException() throws Excepti
...
+    assertEquals("", placesFetcher.createCuisinesQuery(ImmutableList.of()));
+  }
+
+  @Test
+  public void createCuisinesQuery_getsInvalidCuisines_throwsFetcherException() throws Exception {
+    FetcherException thrown =
+        assertThrows(
+            FetcherException.class,
+            () -> placesFetcher.createCuisinesQuery(ImmutableList.of("blah")));
+    assertTrue(thrown.getCause() instanceof NullPointerException);
```
arielz-g, line 236: Shouldn't this be a FetcherException?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -25,27 +26,38 @@
...
+    // The entry point for a Google GEO API request.
     private static final GeoApiContext CONTEXT = new GeoApiContext.Builder()
         .apiKey(System.getenv("API_KEY"))
         .build();
 
+    // The path of the configuration file containing the mapping of cuisines to search words.
+    private static final String CUISINES_SEARCH_WORDS_CONFIG_PATH  = "cuisinesSearchWords.json";
+
+    // A mapping between cuisines and text search words. */
+    private static final Map<String, List<String>> CUISINE_TO_SEARCH_WORDS = getCuisinesMap();
```
arielz-g: Should change here to ImmutableMap as well.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: The exception thrown is a FetcherException. Each FetcherException has a cause, so in this case of throwing fetcherException due to invalid cuisine the cause of the exception will be NullPointerException. Do you think I should change the cause to some more general exception?

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Sure, I changed it

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
arielz-g: Ah my bad, missed that part.

arielz-g (approved):

PR #35, opened 2020-11-09
Send user input to fetcher

karenav:
Hey, I would love to hear your thoughts especially about the tests - if you think more tests are required, due to passing the user input to the fetch.

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -22,8 +22,11 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.data.Places;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.gson.Gson;
+import com.google.maps.model.LatLng;
 import com.google.sps.data.PlacesFetcher;
+import com.google.sps.data.UserPrefrences;
 
 /**
  * A servlet that handles the user query. Currently accepts no input, and responds with a list of
@@ -49,10 +52,18 @@ public final class QueryServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     try {
+      UserPrefrences userPrefs =
+          UserPrefrences.builder()
+              .setMinRating(Float.parseFloat(request.getParameter("rating")))
+              .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
+              .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
+              .setLocation(getLatLngFromString(request.getParameter("location")))
+              .setCuisineTypes(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
+              .build();
       //TODO(M1): add call to filterer
       response.setContentType("application/json");
       response.getWriter().write(new Gson().toJson(
-          Places.randomSort(fetcher.fetch())
+          Places.randomSort(fetcher.fetch(userPrefs))
           .stream()
           .limit(MAX_NUM_PLACES_TO_RECOMMEND)
           .collect(Collectors.toList())
@@ -61,4 +72,9 @@ public final class QueryServlet extends HttpServlet {
       //TODO(M1): handle errors - TBD
     }
   }
-}
+
+  private LatLng getLatLngFromString(String coordinates) {
+    String[] latLng = coordinates.split(",");
+    return new LatLng(Float.parseFloat(latLng[0]), Float.parseFloat(latLng[1]));
+  }
+ }
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -71,9 +71,9 @@ public final class QueryServlet extends HttpServlet {
       response.setContentType("application/json");
       response.getWriter().write(new Gson().toJson(
           Places.randomSort(filteredPlaces)
-          .stream()
-          .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-          .collect(Collectors.toList())
+              .stream()
+              .limit(MAX_NUM_PLACES_TO_RECOMMEND)
+              .collect(Collectors.toList())
       ));
     } catch (Exception e) {
       response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -53,6 +53,7 @@ public final class QueryServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    ImmutableList<Place> filteredPlaces;
     try {
       UserPreferences userPrefs =
           UserPreferences.builder()
@@ -60,25 +61,26 @@ public final class QueryServlet extends HttpServlet {
               .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
               .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
               .setLocation(getLatLngFromString(request.getParameter("location")))
-              .setCuisines(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
+              .setCuisines(ImmutableList.copyOf(request.getParameter("cuisines").split(",")))
               .build();
-      ImmutableList<Place> filteredPlaces = Places.filter(
+      filteredPlaces = Places.filter(
           fetcher.fetch(userPrefs) /* places */,
           Integer.parseInt(request.getParameter("rating")) /* min rating */,
           true /* filter if no website */,
           true /* filter branches of same place */
       );
-      response.setContentType("application/json");
-      response.getWriter().write(new Gson().toJson(
-          Places.randomSort(filteredPlaces)
-              .stream()
-              .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-              .collect(Collectors.toList())
-      ));
     } catch (Exception e) {
       response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
           "Fetching from Google Places API encountered a problem");
+      return;
     }
+    response.setContentType("application/json");
+    response.getWriter().write(new Gson().toJson(
+        Places.randomSort(filteredPlaces)
+            .stream()
+            .limit(MAX_NUM_PLACES_TO_RECOMMEND)
+            .collect(Collectors.toList())
+    ));
   }
 
   private LatLng getLatLngFromString(String coordinates) {
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -28,7 +28,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.gson.Gson;
 import com.google.maps.model.LatLng;
 import com.google.sps.data.PlacesFetcher;
-import com.google.sps.data.UserPrefrences;
+import com.google.sps.data.UserPreferences;
 
 /**
  * A servlet that handles the user query. Currently accepts no input, and responds with a list of
@@ -54,13 +54,13 @@ public final class QueryServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     try {
-      UserPrefrences userPrefs =
-          UserPrefrences.builder()
+      UserPreferences userPrefs =
+          UserPreferences.builder()
               .setMinRating(Float.parseFloat(request.getParameter("rating")))
               .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
               .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
               .setLocation(getLatLngFromString(request.getParameter("location")))
-              .setCuisineTypes(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
+              .setCuisines(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
               .build();
       ImmutableList<Place> filteredPlaces = Places.filter(
           fetcher.fetch(userPrefs) /* places */,
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -21,6 +21,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.google.sps.data.FetcherException;
 import com.google.sps.data.Place;
 import com.google.sps.data.Places;
 import com.google.common.annotations.VisibleForTesting;
@@ -31,7 +32,7 @@ import com.google.sps.data.PlacesFetcher;
 import com.google.sps.data.UserPreferences;
 
 /**
- * A servlet that handles the user query. Currently accepts no input, and responds with a list of
+ * A servlet that handles the user's food-mood recommendation query, and responds with a list of
  * recommended places (in Json format).
  */
 @WebServlet("/query")
@@ -69,9 +70,13 @@ public final class QueryServlet extends HttpServlet {
           true /* filter if no website */,
           true /* filter branches of same place */
       );
-    } catch (Exception e) {
+    } catch (FetcherException e) {
       response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
           "Fetching from Google Places API encountered a problem");
+      return;
+    } catch (IllegalArgumentException e) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+          "Parsing the user preferences encountered a problem");
       return;
     }
     response.setContentType("application/json");
@@ -83,7 +88,7 @@ public final class QueryServlet extends HttpServlet {
     ));
   }
 
-  private LatLng getLatLngFromString(String coordinates) {
+  private static LatLng getLatLngFromString(String coordinates) {
     String[] latLng = coordinates.split(",");
     return new LatLng(Float.parseFloat(latLng[0]), Float.parseFloat(latLng[1]));
   }
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -75,7 +75,7 @@ public final class QueryServlet extends HttpServlet {
           "Fetching from Google Places API encountered a problem");
       return;
     } catch (IllegalArgumentException e) {
-      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST,
           "Parsing the user preferences encountered a problem");
       return;
     }
--- a/project/src/main/java/com/google/sps/servlets/QueryServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/QueryServlet.java
@@ -20,6 +20,8 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+
+import com.google.sps.data.Place;
 import com.google.sps.data.Places;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
@@ -60,16 +62,22 @@ public final class QueryServlet extends HttpServlet {
               .setLocation(getLatLngFromString(request.getParameter("location")))
               .setCuisineTypes(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
               .build();
-      //TODO(M1): add call to filterer
+      ImmutableList<Place> filteredPlaces = Places.filter(
+          fetcher.fetch(userPrefs) /* places */,
+          Integer.parseInt(request.getParameter("rating")) /* min rating */,
+          true /* filter if no website */,
+          true /* filter branches of same place */
+      );
       response.setContentType("application/json");
       response.getWriter().write(new Gson().toJson(
-          Places.randomSort(fetcher.fetch(userPrefs))
+          Places.randomSort(filteredPlaces)
           .stream()
           .limit(MAX_NUM_PLACES_TO_RECOMMEND)
           .collect(Collectors.toList())
       ));
     } catch (Exception e) {
-      //TODO(M1): handle errors - TBD
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+          "Fetching from Google Places API encountered a problem");
     }
   }
 
--- a/project/src/main/webapp/script.js
+++ b/project/src/main/webapp/script.js
@@ -40,8 +40,7 @@ function fetchFromQuery() {
   const placesDiv = document.getElementById('place');
   displayResultsPage();
   const map = createMap();
-  const placesDiv = document.getElementById('place');
-  fetch('/query')
+  fetch('/query?' + params)
       .then(response => response.json())
       .then((places) => {
         places.forEach((singlePlace) => {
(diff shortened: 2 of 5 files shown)


project/src/main/webapp/script.js
```diff
@@ -40,13 +40,24 @@ function fetchFromQuery() {
...
-  });
+  fetch('/query?' + params)
+      .then(response => response.json())
+      .then((places) => {
+        places.forEach((singlePlace) => {
+          placesDiv.appendChild(createPlaceElement(singlePlace));
+          addPlaceMarker(map, singlePlace)
+        });
+        displayAfterResults();
+        if (places.length < 3) {
```
tbarnahor: It is great that you added this handeling. I just want to bring to attention that it doesn't match what we planned in the [design](https://docs.google.com/document/d/1yPTb2Rh6MSTrDE9b9e1Q61ZCcH_cVNqBI8qs1VKNZEs/edit#bookmark=id.hr310z4jfen1).
I am fine with not doing what we said regarding the zero results because it may complicate things and we don't have time for that, but then we should update the design.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -49,16 +54,35 @@ void init(PlacesFetcher inputFetcher) {
...
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     try {
-      //TODO(M1): add call to filterer
+      UserPreferences userPrefs =
+          UserPreferences.builder()
+              .setMinRating(Float.parseFloat(request.getParameter("rating")))
+              .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
+              .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
+              .setLocation(getLatLngFromString(request.getParameter("location")))
+              .setCuisines(ImmutableList.copyOf(request.getParameter("quisines").split(",")))
```
tbarnahor: cuisines

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -49,16 +54,35 @@ void init(PlacesFetcher inputFetcher) {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     try {
```
tbarnahor, line 58: If possible, I think the try block should be more focused on the lines causing the handeled exception.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -95,8 +112,20 @@ public void getRequest_fetchedLessThanMaxNumPlaces_respondAllFetchedPlaces() thr
...
+    System.out.println(responseStringWriter.getBuffer());
+    System.out.println(responseStringWriter.getBuffer().toString());
+    System.out.println(new Gson()
+    .fromJson(responseStringWriter.getBuffer().toString(), JsonArray.class));
     return new Gson()
         .fromJson(responseStringWriter.getBuffer().toString(), JsonArray.class)
         .size();
   }
+
+  private void setRequestParameters(String quisines) {
```
tbarnahor: cuisines

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -95,8 +112,20 @@ public void getRequest_fetchedLessThanMaxNumPlaces_respondAllFetchedPlaces() thr
 
   // Returns the number of json elements in the servlet's response
   private int getPlacesAmountInResponse() {
+    System.out.println(responseStringWriter.getBuffer());
```
tbarnahor: Remove the prints :)

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: I like this solution of putting all the getPramater treatment in one function :)

project/src/main/webapp/script.js
karenav: Yeah, as we recently decided (and updated in the Engineering Requirements part), the more graceful solution will be implemented in P2. This was not updated in the Product Requirements so I updated it accordingly :)

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: That's a great point. This try-catch is in order to send an error to the "fetch" call, so that we'll go to the catch there if there is any error in the servlet (so that the code doesn't break). what to you think is the right scope for the "try"? I'm not sure about it.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: right, thanks!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: absolutely right. done

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: great, done

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Thank you!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: From the catch message you wrote I understand that it handles fetching problems, so probably surround only the fetch call (or maybe all lines 65 - 70). But really whatever you think.

tbarnahor (approved):

tbarnahor (approved):

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -20,10 +20,15 @@
...
 import com.google.sps.data.Places;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.gson.Gson;
+import com.google.maps.model.LatLng;
 import com.google.sps.data.PlacesFetcher;
+import com.google.sps.data.UserPreferences;
 
 /**
  * A servlet that handles the user query. Currently accepts no input, and responds with a list of
```
szwbar: I think this comment needs to be updated.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -48,17 +53,38 @@ void init(PlacesFetcher inputFetcher) {
...
-          .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-          .collect(Collectors.toList())
-      ));
+      UserPreferences userPrefs =
+          UserPreferences.builder()
+              .setMinRating(Float.parseFloat(request.getParameter("rating")))
+              .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
+              .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
+              .setLocation(getLatLngFromString(request.getParameter("location")))
+              .setCuisines(ImmutableList.copyOf(request.getParameter("cuisines").split(",")))
```
szwbar, line 65: Redundant space?

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -48,17 +53,38 @@ void init(PlacesFetcher inputFetcher) {
...
-          Places.randomSort(fetcher.fetch())
-          .stream()
-          .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-          .collect(Collectors.toList())
-      ));
+      UserPreferences userPrefs =
+          UserPreferences.builder()
+              .setMinRating(Float.parseFloat(request.getParameter("rating")))
+              .setMaxPriceLevel(Integer.parseInt(request.getParameter("price")))
+              .setOpenNow(Integer.parseInt(request.getParameter("open")) != 0)
```
szwbar, line 63: Can this parameter be null as well?

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -48,17 +53,38 @@ void init(PlacesFetcher inputFetcher) {
...
+              .setLocation(getLatLngFromString(request.getParameter("location")))
+              .setCuisines(ImmutableList.copyOf(request.getParameter("cuisines").split(",")))
+              .build();
+      filteredPlaces = Places.filter(
+          fetcher.fetch(userPrefs) /* places */,
+          Integer.parseInt(request.getParameter("rating")) /* min rating */,
+          true /* filter if no website */,
+          true /* filter branches of same place */
+      );
     } catch (Exception e) {
```
szwbar: I think you need to catch specific Exceptions and handle them accordingly. 
For example, the Fetcher throws a `FetcherException`, which means Google maps API had a problem. 
But you might end up with NullpointerException, RuntimeException, or many other exceptions. 
So, first catch what you know might be thrown, and then handle this in a specific way (here, returning an error with the message "Fetching from Google Places API encountered a problem"). 
Then, catch all other exceptions and send the relevant error (which in this case, won't be fetching from google places API).

szwbar: Another option will be to catch only what you intend that will be thrown. Everything else will be returned as server error, which you can then deal with in the Frontend.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -48,17 +53,38 @@ void init(PlacesFetcher inputFetcher) {
...
+    response.setContentType("application/json");
+    response.getWriter().write(new Gson().toJson(
+        Places.randomSort(filteredPlaces)
+            .stream()
+            .limit(MAX_NUM_PLACES_TO_RECOMMEND)
+            .collect(Collectors.toList())
+    ));
+  }
+
+  private LatLng getLatLngFromString(String coordinates) {
```
szwbar: static?

project/src/main/webapp/script.js
```diff
@@ -40,13 +39,24 @@ function fetchFromQuery() {
...
+          document.getElementById('message-container').innerHTML =
+              'Your search had ' + places.length + ' results. You are welcome to try again, \
```
szwbar: "only" X results ?

project/src/main/webapp/script.js
```diff
@@ -40,13 +39,24 @@ function fetchFromQuery() {
...
+              and maybe try to change some of the entered parameters.'
+        }
+      })
+      .catch((error) => {
```
szwbar, line 60: 👍

project/src/main/webapp/script.js
```diff
@@ -220,7 +228,7 @@ function createMapMarker(map, placePosition, placeTitle) {
   })
 }
 
-/** Displays a Google Maps map that allows the user to search fo his location. */
+/** Displays a Google Maps map that allows the user to search for his location. */
```
szwbar: for his location => for their location

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -54,9 +56,10 @@ public void setUp() throws Exception {
 
   @Test
   public void getRequest_fetchedMoreThanMaxNumPlaces_respondMaxNumPlaces() throws Exception {
+    setRequestParameters(/*cuisines*/ "sushi");
```
szwbar: Suggestion: add an overload to `setRequestParameters`, which does not accept any parameters. This will initialize a request with a default set ({"sushi"}).

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -99,4 +116,14 @@ private int getPlacesAmountInResponse() {
         .fromJson(responseStringWriter.getBuffer().toString(), JsonArray.class)
         .size();
   }
+
+  // Set the parameters for the mocked http request. Most parameter are contant valid values, and
+  // the cuisines parameter is supplied by the calling function.
+  private void setRequestParameters(String cuisines) {
```
szwbar: Suggestion: rename to `initializeRequestParameters`. Seems clearer to me as you don't really set anything..

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -15,6 +15,7 @@
 package com.google.sps.servlets;
```
szwbar, line 15: I'm missing multiple tests in this file:
1. Test that the parameters given in the request are actually the parameters that are inserted into the UserPreferences struct. 
2. Test error handling: what happens if the fetcher throws an exception? the UserPreferencesBuilder? The sorter? JSON parsing error? All of those should be tested (separately!).
3. Test that filtering was actually performed according to the given values. The filter method is static, so you cannot mock it, but you should know what are its results - so make sure it filters correctly.

szwbar (changes requested):
Generally LGTM, mostly missing tests.

project/src/main/webapp/script.js
karenav: Hahaha I also thought about it, but because it also applies to 0 I thought it wouldn't fit. 
Do you think we should divide to different cases? I can do that

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: where?

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: I'm not sure I understand your question, could you please explain? 
always 0 or 1 are sent to the parameter, it can never receive nothing. are you asking if we can change it so that it'll be null when the user doesn't mind?

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Great, thanks. 
The reason you think it should be static is that it doesn't use any class variables and that it has a small and a "utility like" job? (I'm asking to acquire intuition for deciding that)

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: done, thanks

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: got it, thanks! Tried to add a more accurate error handling

project/src/main/webapp/script.js
karenav: Done, thanks. Changed in more places as well.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Great idea, thanks

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: That's a great point. 
I decided to handle the whole thing a bit differently, in a way that I think is now better.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Added a few tests that I hope will answer the requirements. 
About the first test - I'm not sure how such a test is supposed to look, it would be great if we could discuss that

project/src/main/java/com/google/sps/servlets/QueryServlet.java
szwbar: Hm, I guess something was misleading in the GitHub UI. Sorry!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -69,10 +70,14 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
           true /* filter if no website */,
           true /* filter branches of same place */
       );
-    } catch (Exception e) {
+    } catch (FetcherException e) {
       response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
           "Fetching from Google Places API encountered a problem");
       return;
+    } catch (IllegalArgumentException e) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
```
szwbar: This is most definitely not an `SC_INTERNAL_SERVER_ERROR`, but rather an `SC_BAD_REQUEST`. 

When returning an HTTP response, it's important to distinguish between server and client error. 
A server error is your way of telling the client: "I had a problem, my bad!". The client can go ahead and retry in a short while, and potentially get a response. These are all represented as 500 errors. The sub errors (5XX) represent more nuanced view into what happened. For instance, 503 says that server is unavailable (which leads to network/setup/etc. problems), while 501 (NOT_IMPLEMENTED) tells the client that this functionality does not exist in the server (so it shouldn't retry the reuqest!)

A client error is when the client did something wrong. Like - provide wrong arguments. Retrying the request won't do any good here! They should look into what they did wrong and fix the problem.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -91,39 +92,89 @@ public void getRequest_fetchMoreThanOneCuisine_success() throws Exception {
     assertEquals(getPlacesAmountInResponse(), QueryServlet.MAX_NUM_PLACES_TO_RECOMMEND);
   }
 
+  @Test
+  public void getRequest_fetchedPlacesShouldBeFiltered_filterPlaces() throws Exception {
+    when(REQUEST.getParameter("rating")).thenReturn("4");
+    Place validPlace = createValidPlaceBuilder().setName("validPlace").build();
+    Place lowRating = createValidPlaceBuilder().setRating(3).setName("lowRatingPlace").build();
+    Place noWebsite = createValidPlaceBuilder().setWebsiteUrl("").setName("noWebsitePlace").build();
+    // The following place purposely has the same name as the valid place, so that we make sure that
```
szwbar, line 88: Great comment!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -91,39 +92,89 @@ public void getRequest_fetchMoreThanOneCuisine_success() throws Exception {
...
+    // we end up with only one of them.
+    Place anotherBranchOfValidPlace = createValidPlaceBuilder().setName("validPlace").build();
```
szwbar: Suggestion: maybe rename to validPlace2?

project/src/main/java/com/google/sps/servlets/QueryServlet.java
szwbar: Correct! I think it doesn't have any logic, but it's a small utility class.

project/src/main/webapp/script.js
szwbar: I think that two different strings is the way to go.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
szwbar: Instead of calling 
`when(FETCHER.fetch(any(UserPreferences.class))).`
you need to build the expected UserPreferences object that your class should build given the input you give it in the test, and see that everything is populated correctly. 

Something like 
```java
UserPreferences expectedUP = UserPreferences.builder().setFoo(...)...build();
when(FETCHER.fetch(expectedUp)).then(...); 
```
And then use `verify` to verify the call has actually happened. 

Another option is still use the `any(UserFeedback.class)` and only test the `expectedUP` was provided by calling `verify` on your mock later.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
szwbar: Looks good!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
szwbar: "never" is a strong word. 
You rely on the client to always provide the right input - but this means you will fail if it doesn't! 
For instance, if someone changes the client and it no longer sends this field, it will be null! This can always and does happen, so your code should be immune and not fail ungracefully.

szwbar (changes requested):
Great work! I think one more test should be added and you'r good to go. 

Really liked the tests you wrote, very concise and clear.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: Sorry, I still don't understand your question. 
All the other lines can be broken as well if we don't get them from the client.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: That's a great point, thanks for clearing that. Changed.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: done

project/src/main/webapp/script.js

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Added the test, thanks for the help. 
As we discussed, it will be tested when we refactor the float to be a double, because the float is unstable.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
szwbar: Actually, you're right. All of those will throw some exception... So you can leave as-is, I guess.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
szwbar: 👍

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -67,36 +72,122 @@ public void getRequest_fetchedMoreThanMaxNumPlaces_respondMaxNumPlaces() throws
...
+        .setMinRating(4)
+        .setMaxPriceLevel(3)
+        .setOpenNow(true)
+        .setLocation(new LatLng(30.30000000, 35.35000000))
+        .setCuisines(ImmutableList.of("sushi", "hamburger"))
+        .build();
+
+    servlet.doGet(REQUEST, RESPONSE);
+
+    // TODO(M1): this test fails because float is unstable for that kind of a test.
```
szwbar: I prefer that you simply make a simple float like "0". Doesn't this work? 
And then the TODO will be to change the float.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -67,36 +72,122 @@ public void getRequest_fetchedMoreThanMaxNumPlaces_respondMaxNumPlaces() throws
...
+
+  @Test
+  // This test checks that the user preferences are processed correctly based on the given
+  // parameters. The crucial point of the test is that the strings that are given as parameters
+  // match the expect user preferences.
+  public void getRequest_userPreferencesForwardedToFetcher() throws Exception {
+    when(REQUEST.getParameter("rating")).thenReturn("4");
+    when(REQUEST.getParameter("price")).thenReturn("3");
+    when(REQUEST.getParameter("open")).thenReturn("1");
+    when(REQUEST.getParameter("location")).thenReturn("30.30000000,35.35000000");
```
szwbar: Why don't you use `initializeRequestParameters`? 
Using global constants can help you build the UserPreferences according to the values. 

Also, you don't seem to call `when` on the mock. Doesn't this cause problems in the test?

szwbar (approved):
Generally approved, only comments on the last test. 
Great work!!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: I thought it's best to do it inside the function, because if someone changes something in initializeRequestParameters (in a naive manner) it might break the test. And it relates for me with what Ariel says - that in tests we give a high value to readability, so I thought that's the clearer way to show it.  What do you think?
I'm not sure what you're asking regarding the mock, maybe we should discuss it f2f if you think it's an issue?

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: Great idea, this does work. done.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
szwbar: Ack.

PR #36, opened 2020-11-10
Return more details from placesFetcher and add them to Place

tbarnahor:


project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -40,35 +40,50 @@
...
   private static final Place PLACE_2 =
-      Place.builder()
-      .setName("name2")
-      .setWebsiteUrl(PLACE_URL)
-      .setPhone(PHONE)
-      .setRating(RATING)
-      .setPriceLevel(PRICELEVEL_INT)
-      .setLocation(LOCATION)
-      .build();
+  Place.builder()
```
karenav: I think it should be indented

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -109,15 +124,19 @@ private static URL createTestURL(String s) {
...
     PlaceDetails placeDetails = new PlaceDetails();
     placeDetails.name = name;
-    placeDetails.website = url;
+    placeDetails.website = website;
     placeDetails.formattedPhoneNumber = phone;
     placeDetails.rating = rating;
     placeDetails.priceLevel = priceLevel;
     placeDetails.geometry = new Geometry();
     placeDetails.geometry.location = location;
+    placeDetails.url = url;
```
karenav: If this is the google URL maybe it should be clearer (adding "google")?

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -29,6 +33,10 @@
     private static final int MAX_PRICE_LEVEL = 4;
     /** The minimal valid price level value. */
     private static final int MIN_PRICE_LEVEL = 0;
+    /** The optional business statuses. */
+    private static final List<String> OPTIONAL_STATUS =
```
karenav: I was confused by the name, I think maybe something like "valid statuses" would be more clear at least for me.

karenav (approved):
Looks really great!

project/src/main/java/com/google/sps/data/ValidationUtils.java
tbarnahor: You are right, valid is much better.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Those are the fields of the google places API (of the PlaceDetails type), I agree it is very confusing so I changed the function parameter like you suggested.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Totaly true

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -54,9 +54,24 @@
...
+   */
+  public abstract String googleUrl();
+
+  /**
+    * @return the place's placeId as defiened on Google Places.
+    */
+  public abstract String placeId();
+
+  /**
+    * @return the place's business status
```
szwbar: Please document here as well the possible values for businessStatus.

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -110,6 +125,26 @@ public static Builder builder() {
...
+     */
+    public abstract Builder setPlaceId(String placeId);
+
+    /**
+     * @param businessStatus the place's status, represented by one of the following Strings:
+     * "OPERATIONAL", "CLOSED_TEMPORARILY", "CLOSED_PERMANENTLY" and null if the information
+     * isn't available.
+     * @return a Place builder that enables to continue building
+     */
+    public abstract Builder setBusinessStatus(String businessStatus);
```
szwbar: Have you considered making this an `enum` instead of String? This way you'll always have one of multiple available values, and won't run into any surprises and won't need to do any validation.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -108,6 +108,10 @@
                     .setRating(placeDetails.rating)
                     .setPriceLevel(Integer.parseInt(placeDetails.priceLevel.toString()))
                     .setLocation(placeDetails.geometry.location)
+                    .setPlaceId(placeDetails.placeId)
+                    .setGoogleUrl(placeDetails.url == null
+                            ? "" : placeDetails.url.toString())
```
szwbar: Guava has a `nullToEmpty` method which is very useful: https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html

There are many more useful methods there, so maybe you should add Guava to your project and use it for String manipulation.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -40,34 +40,49 @@
 @RunWith(JUnit4.class)
 public final class PlacesFetcherTest {
 
-  private static final String PLACE_URL = "https://www.google.com/"; // used for Places
-  private static final URL PLACES_DETAILS_URL = createTestURL(PLACE_URL); // used for PlaceDetails
+  private static final String PLACE_WEBSITE = "https://www.google.com/"; // used for Places
+  private static final String PLACE_GOOGLE_URL = "https://www.google.com/"; // used for Places
```
szwbar, line 46: Make this a different string. This way, if you ended up mixing the two strings, the test will find that out.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -40,34 +40,49 @@
...
+  private static final URL PLACE_DETAILS_WEBSITE =
+      createTestURL(PLACE_WEBSITE); // used for PlaceDetails
+  private static final URL PLACE_DETAILS_GOOGLE_URL =
+      createTestURL(PLACE_GOOGLE_URL); // used for PlaceDetails
   private static final String PHONE = "+97250-0000-000";
   private static final LatLng LOCATION = new LatLng(32.08074, 34.78059);
   private static final float RATING = 4;
   private static final int PRICELEVEL_INT = 2; // used for Places and UserPreferences
   private static final PriceLevel PRICELEVEL = PriceLevel.MODERATE; // used for PlaceDetails
```
szwbar: Note in your PR, but this should be PRICE_LEVEL and not PRICELEVEL.

szwbar (changes requested):
Looks good!

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Changed it

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Right.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Great, I used this where possible. I also found a better solution for the URL to String conversion that wasn't elegant before because of the null case.

project/src/main/java/com/google/sps/data/Place.java
tbarnahor: Changed it to enum, it does handel the business status better. I don't have exprience with using enums so let me know if this is really the way to do it.

project/src/main/java/com/google/sps/data/Place.java
tbarnahor: Added

project/src/main/java/com/google/sps/data/BusinessStatus.java
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Used by Places builder to restrict the buisiness statuses
+ *  to match those returned by Google Places API.
+ */
+public enum BusinessStatus  {
```
szwbar, line 20: Looks good!

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -54,9 +54,25 @@
...
+   */
+  public abstract String googleUrl();
+
+  /**
+    * @return the place's placeId as defiened on Google Places.
+    */
+  public abstract String placeId();
+
+  /**
+    * @return the place's business status, represented by one of the following:
```
szwbar: Now that it's an enum, you can remove the documentation as it's clear from the enum definition.

szwbar (approved):

project/src/main/java/com/google/sps/data/Place.java
karenav: I have a semi technical question - can an enum be in a class that's not named after it? 
I that's possible, maybe we should put the enum in the util class for Place - Places.java?

project/src/main/java/com/google/sps/data/Test.java
```diff
@@ -0,0 +1,26 @@
+package com.google.sps.data;
```
karenav: Is this class supposed to be here?

project/src/main/java/com/google/sps/data/Place.java
tbarnahor: Yes it is possible, it can be defined on the Places class for example and then approached by Places.BusinessStatus 
I saw online that the best practice for public enums is to define them on a seperate file since it is just like a class, and that makes sense to me.

szwbar: I agree, this way it can be shared among different modules in the future, without them needing to know some util file.

project/src/main/java/com/google/sps/data/Test.java
szwbar: +1 ?

PR #38, opened 2020-11-23
turn float parse to double parse

karenav:
This is a tiny PR, meant for improving the query servlet - using double instead of float.

### DIFF
--- a/project/src/test/java/com/google/sps/servlets/QueryServletTest.java
+++ b/project/src/test/java/com/google/sps/servlets/QueryServletTest.java
@@ -134,20 +134,17 @@ public final class QueryServletTest {
     when(REQUEST.getParameter("rating")).thenReturn("4");
     when(REQUEST.getParameter("price")).thenReturn("3");
     when(REQUEST.getParameter("open")).thenReturn("1");
-    when(REQUEST.getParameter("location")).thenReturn("00.00000000,00.00000000");
+    when(REQUEST.getParameter("location")).thenReturn("35.35000000,30.00000000");
     when(REQUEST.getParameter("cuisines")).thenReturn("sushi,hamburger");
     UserPreferences expectedUserPrefs = UserPreferences.builder()
         .setMinRating(4)
         .setMaxPriceLevel(3)
         .setOpenNow(true)
-        .setLocation(new LatLng(00, 00))
+        .setLocation(new LatLng(35.35000000, 30.00000000))
         .setCuisines(ImmutableList.of("sushi", "hamburger"))
         .build();
 
     servlet.doGet(REQUEST, RESPONSE);
-
-    // TODO(M1): this test fails with semi-real ccordinates, because float is unstable.
-    // When we refactor the float to be a double, we'll need to add real values (instead of 0).
 
     verify(FETCHER).fetch(expectedUserPrefs);
   }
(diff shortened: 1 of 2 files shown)


project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -134,21 +134,18 @@ public void getRequest_userPreferencesForwardedToFetcher() throws Exception {
     when(REQUEST.getParameter("rating")).thenReturn("4");
     when(REQUEST.getParameter("price")).thenReturn("3");
     when(REQUEST.getParameter("open")).thenReturn("1");
-    when(REQUEST.getParameter("location")).thenReturn("00.00000000,00.00000000");
+    when(REQUEST.getParameter("location")).thenReturn("35.35000000,30.00000000");
```
tbarnahor, line 137: Are all those zeros necessary? What happens if you just write .thenReturn("35.35,30")?

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -134,21 +134,18 @@ public void getRequest_userPreferencesForwardedToFetcher() throws Exception {
...
     when(REQUEST.getParameter("cuisines")).thenReturn("sushi,hamburger");
     UserPreferences expectedUserPrefs = UserPreferences.builder()
         .setMinRating(4)
         .setMaxPriceLevel(3)
         .setOpenNow(true)
-        .setLocation(new LatLng(00, 00))
+        .setLocation(new LatLng(35.35000000, 30.00000000))
```
tbarnahor, line 143: Are all those zeros necessary? What happens if you just write "new LatLng(35.35, 30)?

karenav: same...

karenav: It's just to make it clear that it matches the previous latlng. No difference in the functionality

tbarnahor: Why does the previous latlng need all those zeros? The varification doesn't match without them?

tbarnahor (approved):

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
karenav: It doesn't "need" them, it's the same functionality with and without :) 
The previous one looks that way because it resembles how the user input is passed on as a parameter (obviously usually it's not zeroes)

karenav: (I don't mind changing it if it bothers you)

tbarnahor: I don't really care, I was just curious for the reason to write it this way

szwbar: I actually agree with Tal here. I don't think this improves readability. 

Consider giving a really small fraction (e.g. 35.000005) and verify it works correctly. The way it's structured now, small differences in the results (such as the one that failed this test originally) aren't captured.

szwbar (approved):
LGTM!

PR #39, opened 2020-11-23
Context singelton

tbarnahor:
A small PR to add the geo contex as a singleton class to be used on placesFetcher and placesSorter

### DIFF
--- a/project/src/main/java/com/google/sps/data/GeoContext.java
+++ b/project/src/main/java/com/google/sps/data/GeoContext.java
@@ -1,0 +1,52 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.maps.GeoApiContext;
+
+/**
+ * GeoContext holds the entry point for a Google GEO API request.
+ * It is initiated once and used for all Google GEO EPI requests,
+ * including Google Places API and Google Distance Matrix API.
+ *
+ */
+
+public class GeoContext {
+
+    private static GeoContext singleInstance = null;
+
+    private GeoApiContext context;
+
+    private GeoContext() {
+        context = new GeoApiContext.Builder()
+            .apiKey(System.getenv("API_KEY"))
+            .build();
+    }
+
+    /**
+     * Creates a GeoContext instance if not yet been anitialized and returns the
+     * intance's contex. The double check of the instance being null is for thread safety.
+     */
+    public static GeoApiContext getContext() {
+        if (singleInstance == null){
+            synchronized(GeoContext.class) {
+                if (singleInstance == null) {
+                    singleInstance = new GeoContext();
+                }
+            }
+        }
+        return singleInstance.context;
+    }
+}
--- a/project/src/main/java/com/google/sps/data/GeoContext.java
+++ b/project/src/main/java/com/google/sps/data/GeoContext.java
@@ -34,12 +34,12 @@ public final class GeoContext {
     }
 
     /**
-     * Creates a GeoContext instance if not yet been anitialized and returns the
-     * intance's contex. The double check of the instance being null is for thread safety.
+     * Creates a GeoContext instance if not yet been anitialized and returns the intance's context.
      *
      * @return a GeoApiContext object used for Google GEO API requests
      */
-    public static GeoApiContext getContext() {
+    public static GeoApiContext getGeoApiContext() {
+        // Double check if initialized for thread safety.
         if (singleInstance == null) {
             synchronized (GeoContext.class) {
                 if (singleInstance == null) {
--- a/project/src/main/java/com/google/sps/data/GeoContext.java
+++ b/project/src/main/java/com/google/sps/data/GeoContext.java
@@ -23,7 +23,7 @@ import com.google.maps.GeoApiContext;
  *
  */
 
-public class GeoContext {
+public final class GeoContext {
 
     private static GeoContext singleInstance = null;
 
@@ -38,10 +38,11 @@ public class GeoContext {
     /**
      * Creates a GeoContext instance if not yet been anitialized and returns the
      * intance's contex. The double check of the instance being null is for thread safety.
+     * @return a GeoApiContext object used for Google GEO API requests
      */
     public static GeoApiContext getContext() {
-        if (singleInstance == null){
-            synchronized(GeoContext.class) {
+        if (singleInstance == null) {
+            synchronized (GeoContext.class) {
                 if (singleInstance == null) {
                     singleInstance = new GeoContext();
                 }
--- a/project/src/main/java/com/google/sps/data/GeoContext.java
+++ b/project/src/main/java/com/google/sps/data/GeoContext.java
@@ -23,30 +23,24 @@ import com.google.maps.GeoApiContext;
  */
 public final class GeoContext {
 
-    private static GeoContext singleInstance = null;
-
-    private GeoApiContext context;
-
-    private GeoContext() {
-        context = new GeoApiContext.Builder()
-            .apiKey(System.getenv("API_KEY"))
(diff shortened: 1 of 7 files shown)


project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.google.maps.GeoApiContext;
+
+/**
+ * GeoContext holds the entry point for a Google GEO API request.
+ * It is initiated once and used for all Google GEO EPI requests,
+ * including Google Places API and Google Distance Matrix API.
+ *
+ */
+
```
karenav: nit: extra space

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public static GeoApiContext getContext() {
+        if (singleInstance == null) {
+            synchronized (GeoContext.class) {
+                if (singleInstance == null) {
+                    singleInstance = new GeoContext();
+                }
+            }
+        }
+        return singleInstance.context;
```
karenav: could this fail in earlier stages? there is no way this will try to get the "context" of a null object? or an error will be thrown if an earlier stage fails?

karenav (approved):

project/src/main/java/com/google/sps/data/GeoContext.java
tbarnahor: That's a good point. This shouldn't fail because if there is any problem with the API key the GeoApiContext will still be built with no errors. The errors will come only when the context is used by a request. Actualy I found out that this case (of invalid API key) wasn't treated in the fetcher so I added it there.

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,52 @@
...
+    private GeoApiContext context;
+
+    private GeoContext() {
+        context = new GeoApiContext.Builder()
+            .apiKey(System.getenv("API_KEY"))
+            .build();
+    }
+
+    /**
+     * Creates a GeoContext instance if not yet been anitialized and returns the
```
szwbar: typo: initialized

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,52 @@
...
+     * intance's contex. The double check of the instance being null is for thread safety.
```
szwbar: typo: context

szwbar: > The double check of the instance being null is for thread safety

Not part of the API, shouldn't be part of the javadoc. Move this comment to the function itself.

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,52 @@
+// Copyright 2019 Google LLC
```
szwbar, line 1: Add a test for this file?

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,52 @@
...
+            .build();
+    }
+
+    /**
+     * Creates a GeoContext instance if not yet been anitialized and returns the
+     * intance's contex. The double check of the instance being null is for thread safety.
+     *
+     * @return a GeoApiContext object used for Google GEO API requests
+     */
+    public static GeoApiContext getContext() {
```
szwbar: `GetContext` is a bit confusing to me, since you have this class called `GeoContext` and also `GeoApiContext`. Maybe change to `getGeoApiContext`?

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,52 @@
...
+package com.google.sps.data;
+
+import com.google.maps.GeoApiContext;
+
+/**
+ * GeoContext holds the entry point for a Google GEO API request.
+ * It is initiated once and used for all Google GEO EPI requests,
+ * including Google Places API and Google Distance Matrix API.
+ */
+public final class GeoContext {
```
szwbar, line 24: I think there's one layer too many here. 

Why do you need an instance of `GeoContext` to hold the `GeoApiContext`? You can have the `GeoApiContext` as the `singleInstance` directly, and thus simplify this class.

WDYT?

project/src/main/java/com/google/sps/data/GeoContext.java
tbarnahor: Right

project/src/main/java/com/google/sps/data/GeoContext.java
tbarnahor: You are right

project/src/main/java/com/google/sps/data/GeoContext.java
tbarnahor: Agreed, I changed the class a bit to get rid of that layer. Let me know if that's what you meant.

project/src/main/java/com/google/sps/data/GeoContext.java
tbarnahor: Added a test

tbarnahor:
I also changed the call to the GeoContext to be from the Servlet instead of from the fetcher (dependency injection).

project/src/main/java/com/google/sps/data/GeoContext.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Creates a GeoApiContext instance if not yet been initialized and returns it.
+     *
+     * @return a GeoApiContext object used for Google GEO API requests
+     */
+    public static GeoApiContext getGeoApiContext() {
+        // Double check if initialized for thread safety.
+        if (singleContextInstance == null) {
+            synchronized (GeoContext.class) {
```
szwbar, line 36: Super interesting! Didn't know it works this way.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -42,10 +44,12 @@
   static final int MAX_NUM_PLACES_TO_RECOMMEND = 3;
   private PlacesFetcher fetcher;
 
+    // The entry point for a Google GEO API request.
+  private static final GeoApiContext CONTEXT = GeoContext.getGeoApiContext();
```
szwbar: Now that you have a singleton factory, you don't really need to hold it as a static member of your own. 
You can just have `new PlacesFetcher(GeoContext.getGeoApiContext());`

If you need many, just store it as a temporary variable in the `init()` function.

project/src/test/java/com/google/sps/data/GeoContextTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+
+@RunWith(JUnit4.class)
+public class GeoContextTest {
+
+    @Test
+    public void getGeoApiContext_followingCalls_returnSameGeoApiContext() {
```
szwbar: nit: subsequent calls

project/src/test/java/com/google/sps/data/GeoContextTest.java
```diff
@@ -0,0 +1,31 @@
...
+      assertEquals(GeoContext.getGeoApiContext(), GeoContext.getGeoApiContext());
```
szwbar, line 30: What if the returned object is null?

szwbar (approved):

project/src/test/java/com/google/sps/data/GeoContextTest.java
tbarnahor: Then it would be null on both calls and the test passes (there is no reason for it ever to be null, but lets say there is some kind of problem with the build that returns null - the subsequent call will result the same, and if not then it's fine the test doen't pass but it should never happen).

szwbar: This is what I meant. If your creation is problematic, this test effectively tests nothing.

szwbar: Testing non-null should suffice here to avoid this problem.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: Totaly right

project/src/test/java/com/google/sps/data/GeoContextTest.java
tbarnahor: Ohhh yes you are right. Added the non-null assertion

PR #40, opened 2020-11-24
Places scorer with hardcoded durations

tbarnahor:
This PR includes the scoreSort as described on [M2](https://docs.google.com/document/d/1yPTb2Rh6MSTrDE9b9e1Q61ZCcH_cVNqBI8qs1VKNZEs/edit#heading=h.hd29yka62dsy), with drivingETAs temporarily hardcoded.
On the next PR the use of distance matrix API will be added, which will complete M2.


project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -45,6 +47,23 @@
...
+  public static ImmutableList<Place> scoreSort(ImmutableList<Place> places, LatLng userLocation) {
+    List<Place> mutablePlaces = new ArrayList<>(places);
+    Map<Place, Double> placesScores = new PlacesScorer(places, userLocation).getScores();
+    Collections.sort(mutablePlaces, (p1, p2) -> {
+      return placesScores.get(p2).compareTo(placesScores.get(p1));
+    });
+    return ImmutableList.copyOf(mutablePlaces);
+  }
+
+
```
karenav, line 65: Looks really elegant!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -81,10 +82,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     response.setContentType("application/json");
     response.getWriter().write(new Gson().toJson(
-        Places.randomSort(filteredPlaces)
-            .stream()
-            .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-            .collect(Collectors.toList())
+      Places.scoreSort(filteredPlaces, userPrefs.location())
+          .stream()
+          .limit(MAX_NUM_PLACES_TO_RECOMMEND)
+          .collect(Collectors.toList())
```
karenav, line 91: Looks good! 
We should think about whether we want to delete the random sort, or we think e might still want to use it. For example, is there an error that the scorer might throw? Maybe when (and if) we catch it we'll want to sort randomly?

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import com.google.maps.model.LatLng;
+
+public class PlacesScorer {
+
+    private static final double RATING_WEIGHT = 0.7;
+    private static final double DURATION_WEIGHT = 0.3;
+    private static final double MAX_RATING = 5;
+    private static final double MAX_DURATION_SECONDS = 60 * 60;
```
karenav: Isn't this a big amount of time? We wouldn't want something smaller?

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
```diff
@@ -0,0 +1,65 @@
...
+        // rating = place's rating / Max Rating
+        // drivingETA = max{1 - durationInMinutes(=30) / 60, 0}
+        Place placeLowRating = PLACE_BUILDER.setName("lowRating").setRating(3).build();
+        Place placeHighRating = PLACE_BUILDER.setName("highRating").setRating(4).build();
+        ImmutableMap<Place, Double> result =
+            new PlacesScorer(ImmutableList.of(placeLowRating, placeHighRating), USER_LOCATION)
+            .getScores();
+        ImmutableMap<Place, Double> expectedResult = ImmutableMap.of(
+            placeLowRating, 0.57,
+            placeHighRating, 0.71);
```
karenav: I think these numbers are a bit confusing. I guess they are the result of a calculation using the constants of the class? I would make the constants VisibleForTesting (package private) and use them in the calculation, instead of writing the result

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -45,6 +47,23 @@
...
   /**
```
karenav, line 66: Do we maybe want a secondary sort in case of a tie? by name or something else?

karenav (approved):

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: I think that since Places API returns us their results in the order that they think we will prefer (as documented on Google Places API) , it is best in case of a tie to preserve the original order.

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
tbarnahor: Yes I wasn't sure about that. I felt that the main thing this test does is to check that there is no unintended change in the scoring equasion (beacuse any change in the algorithm will require a change in test), so I liked the independency. On the other hand, it does look wierd. So I tried to change it a bit, let me know what you think beacuse I am still not sure about it.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: Yeah I guess you are right. I was thinking about it as total delivary time but this is just the driving time so it is too much. I changed it also on the design doc.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: Great idea! I didn't think of that. The scorer might throw an error if there is a problem with the distance matrix API, but I thought in that case to calculate just by rating. I will try to think of more senarios when I add the error handeling, so I think we should keep the randomSort for now.

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: I followed your randomSort structure :)

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -45,6 +47,23 @@
...
   }
 
+  /**
+   * Sorts the given list of Places by score.
+   * @param places the list we want to sort.
+   * @param userLocation the user's loaction, used for scores calculations
+   * @return a new list containing the original list's elements
+   *     arranged by descending order of their scores.
+   */
+  public static ImmutableList<Place> scoreSort(ImmutableList<Place> places, LatLng userLocation) {
```
szwbar: I'm a bit concerned that making this function `static` will be problematic in the future. Will this function need any dependencies? How will you mock them if this function is static? How will we mock this function itself?

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import com.google.maps.model.LatLng;
+
+public class PlacesScorer {
+
+    private static final double RATING_WEIGHT = 0.7;
+    private static final double DURATION_WEIGHT = 0.3;
+    private static final double MAX_RATING = 5;
+    private static final double MAX_DURATION_SECONDS = 40 * 60;
```
szwbar: Please add comments on these fields.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+    private ImmutableMap<Place, Double> durations;
+    private ImmutableList<Place> places;
+
+    /**
+     * Constructor for PlacesScorer, calculates the durations map used for scoring.
+     *
+     * @param placesToScore A list of places we want to calculate their score
+     * @param userLocation The user's physical location used for duration calculations
+     */
+    public PlacesScorer(ImmutableList<Place> placesToScore, LatLng userLocation) {
```
szwbar: Why did you choose to supply these parameters in the c'tor and not in the `getScores` method?

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+        this.durations = getDurations(userLocation);
+    }
+
+   /**
+   * Returns a map of a place and the score the place gets based on a scoring algorithm.
+   * @return A map between a place to a double representing the place’s score
+   */
+    public ImmutableMap<Place, Double> getScores() {
+        Map<Place, Double> scores = new HashMap<>();
+        for (Place place: places) {
```
szwbar: nit: missing whitespace after `:`.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+        this.places = placesToScore;
+        this.durations = getDurations(userLocation);
+    }
+
+   /**
+   * Returns a map of a place and the score the place gets based on a scoring algorithm.
+   * @return A map between a place to a double representing the place’s score
+   */
+    public ImmutableMap<Place, Double> getScores() {
+        Map<Place, Double> scores = new HashMap<>();
```
szwbar: Why create a `HashMap` and then copy it entirely? 

Use `ImmutableMap.Builder` and then `build` it the `return` statement.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private double calcScore(Place place) {
+        return
+            RATING_WEIGHT * (place.rating() / MAX_RATING)
+            + DURATION_WEIGHT * Math.max(1 - (durations.get(place) / MAX_DURATION_SECONDS), 0);
+    }
+
+    // Returns the duration in seconds from each place on places list to the destination
+    private ImmutableMap<Place, Double> getDurations(LatLng destination) {
+        //TODO (M2): This function will call the Distance matrix API to calculate the durations.
```
szwbar: nit: whitespace after `//`.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,59 @@
...
+        //Duration are hardcoded to 30 minutes temporarly.
+        Double duration = 1800D;
+        Map<Place, Double> placesDurations = places.stream()
+            .collect(Collectors.toMap(place -> place, place -> duration));
```
szwbar: I'm pretty sure you can `collect` `toImmutableMap` directly, instead of copying the map's contents.

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
```diff
@@ -0,0 +1,76 @@
...
+        // rating = place's rating / Max Rating
+        // drivingETA = max{1 - durationInMinutes(=30) / 40, 0}
+        Place placeLowRating = PLACE_BUILDER.setName("lowRating").setRating(3).build();
+        Place placeHighRating = PLACE_BUILDER.setName("highRating").setRating(5).build();
+        ImmutableMap<Place, Double> result =
+            new PlacesScorer(ImmutableList.of(placeLowRating, placeHighRating), USER_LOCATION)
+            .getScores();
+        Double expectedScoreLowRatingPlace =
+                RATING_WEIGHT * 3 / MAX_RATING
+                + DURATIONS_WEIGHT * (1 - DURATION_MIN / MAX_DURATION_MIN);
```
szwbar: Instead of making the calculation here, I would make the calculation *in advance* and write the result here. 

You can then leave the calculation as comment, if anyone would like to understand the calculation itself.

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.maps.model.LatLng;
+
+@RunWith(JUnit4.class)
+public class PlacesScorerTest {
+
+    private static final double DELTA = 0.0001;
```
szwbar: Please add a comment on `DELTA`.

szwbar (changes requested):

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: SGTM

project/src/main/java/com/google/sps/data/Places.java
karenav: SGTM

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
tbarnahor: Eventually decided to return the numbers since it looks cleaner.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: Because than I will just need to pass them between all the methods instead of having them as fields and available for all methods. Also, regrading our discussion today I think it makes sense for the durations map to be a field of a PlacesScorer instance because it defines the scores this instance will generate.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: Added

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: No reason, it's the examples I saw online. I changed it to builder.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: Yes I found how to do it now.

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
tbarnahor: I agree

project/src/test/java/com/google/sps/data/PlacesScorerTest.java
tbarnahor: added

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: I will get to this coment next week and will change the Place's functions to be unstatic and also add an interface for the Scorer as we discussed.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Constructor for PlacesScorer, calculates the durations map used for scoring.
+     *
+     * @param placesToScore A list of places we want to calculate their score
+     * @param userLocation The user's physical location used for duration calculations
+     */
+    public PlacesScorer(ImmutableList<Place> placesToScore, LatLng userLocation) {
+        this.places = placesToScore;
+        this.durations = getDurations(userLocation);
```
szwbar: Re-reading this: you're about to be doing substantial work in the constructor. 

This is problematic as constructor cannot throw exceptions, and your object might be left in an uninitialized state.
Consider adding a factory method which will call some private `init()` function of this class, that will call `getDurations()` post construction.

Please read the following:
https://docs.google.com/document/d/1G3XlK4o52agEF9QNYES22wRLI4PJb4NxM-H9KpR0uVc/edit?usp=sharing&resourcekey=0-9boCwlQ22DjV86NOMGWnLA

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,64 @@
...
+    private double calcScore(Place place) {
+        return
+            RATING_WEIGHT * (place.rating() / MAX_RATING)
+            + DURATION_WEIGHT * Math.max(1 - (durations.get(place) / MAX_DURATION_SECONDS), 0);
+    }
+
+    // Returns the duration in seconds from each place on places list to the destination
+    private ImmutableMap<Place, Double> getDurations(LatLng destination) {
+        // TODO (M2): This function will call the Distance matrix API to calculate the durations.
+        //Duration are hardcoded to 30 minutes temporarly.
```
szwbar: nit: whitespace after `//`.

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -45,6 +47,23 @@
...
+  /**
+   * Sorts the given list of Places by score.
+   * @param places the list we want to sort.
+   * @param userLocation the user's loaction, used for scores calculations
+   * @return a new list containing the original list's elements
+   *     arranged by descending order of their scores.
+   */
+  public ImmutableList<Place> scoreSort(ImmutableList<Place> places, LatLng userLocation) {
+    List<Place> mutablePlaces = new ArrayList<>(places);
+    Map<Place, Double> placesScores = new PlacesScorer(places, userLocation).getScores();
```
szwbar: I see you made this function non-`static` but still don't get this as a dependency (or at least some factory, as we discussed).

project/src/main/java/com/google/sps/data/PlacesScorer.java
szwbar: BTW, have a factory is aligned with the design pattern that we talked about yesterday, in which the `Places` class gets a factory of the `Scorer`.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -81,10 +82,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     }
     response.setContentType("application/json");
     response.getWriter().write(new Gson().toJson(
-        Places.randomSort(filteredPlaces)
-            .stream()
-            .limit(MAX_NUM_PLACES_TO_RECOMMEND)
-            .collect(Collectors.toList())
+      Places.scoreSort(filteredPlaces, userPrefs.location())
```
szwbar: How does this work if `scoreSort` is not `static` anymore?

szwbar (changes requested):

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: Changed the scorer to be passed to the scoreSort as an argument, so it is now easily mocked and tested. As for the sorter itself, I don't think it is problematic to leave it static and test it.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: I decided to change it to be in the `getScores` methos since it is easier for testing and also to take the "lazy" approach.

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: I understand and agree with that point. Anyway I decided not to keep durations as a field, but calculate it only when the scorer is called with getScores - it made everything simpler. So currently there is no need in any special scorer construction.

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: Changed it as discussed in above comments.

project/src/main/java/com/google/sps/data/PlacesScorer.java
```diff
@@ -0,0 +1,25 @@
...
+
+package com.google.sps.data;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.maps.model.LatLng;
+
+public interface PlacesScorer {
+
+    ImmutableMap<Place, Double> getScores(ImmutableList<Place> places, LatLng userLocation);
```
szwbar, line 34: Javadoc?

szwbar, line 34: Javadoc?

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    // The maximum durations in seconds,
+    // so that any duration higher than that will not contribute to the place's
+    // score.
+    private static final double MAX_DURATION_SECONDS = 40 * 60;
+
+   /**
+   * Returns a map of a place and the score the place gets based on a scoring algorithm.
+   * @return A map between a place to a double representing the place’s score
+   */
+    @Override
```
szwbar, line 21: Something weird with spacing here....

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        ImmutableMap.Builder<Place, Double> scores = new ImmutableMap.Builder<>();
+        for (Place place : places) {
+            scores.put(place, calcScoreOfPlace(durations, place));
+        }
+        return scores.build();
+    }
+
+    // Calculates a score for place,
+    // score calculated by the place's rating and driving duration from the user's location.
+    private double calcScoreOfPlace(ImmutableMap<Place, Double> durations, Place place) {
```
szwbar: nit: why not full words? 

`calculatePlaceScore`?

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -36,6 +39,21 @@ public void randomSort_keepsAllItems() {
     assertTrue(result.contains(place2));
   }
 
+  @Test
+  public void scoreSort_sortedByScoringAlgorithm() {
+    Place placeLowRating = createValidPlaceBuilderByName("name1").setRating(1).build();
+    Place placeHighRating = createValidPlaceBuilderByName("name2").setRating(2).build();
+    ImmutableList<Place> placesList = ImmutableList.of(placeLowRating, placeHighRating);
+    LatLng userLoaction = new LatLng(32.09, 34.78);
+    PlacesScorer scorerMock = mock(PlacesScorerImpl.class);
```
szwbar: nit: `mockScorer` is the naming convention.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -16,6 +16,7 @@
```
szwbar, line 16: I do think there should be a test here that tests that the scorer was actually called and that the results are sorted. WDYT?

szwbar (approved):
Generally LGTM, small comments left.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: I think that since this only tests the servlet and not the sorter then it should "trust" the sorter that it does its job once its called with the expected parameters. So I added the assetion that the scorer was called with the right parameters. WDYT?

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
tbarnahor: changed it

project/src/main/java/com/google/sps/data/PlacesScorer.java
tbarnahor: Added the javadoc. Since there is no extra information in the javadoc of this function on the Impl class I removed it from there (I understand it inherits this javadoc automaticaly). Is that right?

tbarnahor: I added

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -0,0 +1,49 @@
+package com.google.sps.data;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.maps.model.LatLng;
+
+public class PlacesScorerImpl implements PlacesScorer {
```
karenav, line 7: I feel that this class name doesn't give us more information (and it implies that that's *the* way to implement it). How about something like "PlacesWeightScorer" or "PlacesScorerByWeights"? or something else that adds information about the way the scoring is determined?

tbarnahor: When there is only one implementation, that's the naming convention. If we have more than one implementation (BTW, they will probably all be by some weights) the naming will change.

project/src/main/java/com/google/sps/data/PlacesScorer.java
szwbar: Correct!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
szwbar: LGTM

PR #45, opened 2020-11-24
Filter places according to new Place  properties

### LINKED ISSUE #44: Use Google URL and place status when filtering

karenav:
I updated the filtering function and added tests to check the new filtering requirements

### DIFF
--- a/project/src/test/java/com/google/sps/data/PlacesTest.java
+++ b/project/src/test/java/com/google/sps/data/PlacesTest.java
@@ -17,6 +17,8 @@ package com.google.sps.data;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import java.util.List;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
@@ -72,23 +74,6 @@ public final class PlacesTest {
     assertEquals(result, ImmutableList.of(branch1));
   }
 
-
-  @Test
-  public void filter_noWebsite_filterOut() {
-    Place websiteEmpty = createValidPlaceBuilderByName("name").setWebsiteUrl("").build();
-    Place place1 = createValidPlaceBuilderByName("name1").build();
-    Place place2 = createValidPlaceBuilderByName("name2").build();
-
-    ImmutableList<Place> result = Places.filter(
-        ImmutableList.of(websiteEmpty, place1, place2) /* places */,
-        1 /* min rating */,
-        true /* filter if no website */,
-        false /* filter branches of same place */
-    );
-
-    assertEquals(result, ImmutableList.of(place1, place2));
-  }
-
   @Test
   public void filter_tooLowRating_filterOut() {
     int highRating = 5;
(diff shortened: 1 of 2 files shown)


tbarnahor (approved):
I didn't find anything I would change :)

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -123,6 +106,56 @@ public void filter_filterNotRequired_notFiltering() {
     assertEquals(result, allPlaces);
   }
 
+  @Test
+  public void filter_filterByWebsitePresence() {
```
szwbar, line 124: Can you also add a test for the case when `filterIfNoWebsite` is false? 

It can be simpler than this one.

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -123,6 +106,56 @@ public void filter_filterNotRequired_notFiltering() {
...
+    );
+
+    assertEquals(result, placesToKeep);
+  }
+
+  @Test
+  public void filter_filterByStatus() {
+    Place placeToKeep = createValidPlaceBuilderByName("name1")
+        .setBusinessStatus(BusinessStatus.OPERATIONAL).build();
+    ImmutableList<Place> placesToFilter = ImmutableList.of(
```
szwbar, line 153: Why two places? One is enough to test the functionality IMO.

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -114,7 +97,7 @@ public void filter_filterNotRequired_notFiltering() {
     ImmutableList<Place> allPlaces = ImmutableList.of(websiteEmpty, samePlace1, samePlace2);
 
     ImmutableList<Place> result = Places.filter(
-      allPlaces /* places */,
+        allPlaces /* places */,
```
szwbar: Usually if there's a name to the variable, you don't need to explain what is this variable (i.e. remove the `/* places */ ` comment), it only adds unnecessary explanations to the reader.

szwbar (approved):
Looks great!
A few small comments.

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: Sorry, I'm not sure I understand your question :) 
Each place has a different status, and I wanted to make sure that we keep only the places that have the desired statuses

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: Great point, fixed

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: I had a test that also checks that: filter_filterNotRequired_notFiltering in which filterIfNoWebsite is false, and I made sure that a place that has no site was not filtered. 
I realise maybe it wasn't clear enough that way, so I split it to two separate tests. What do you think?

project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -106,6 +105,21 @@ public void filter_filterNotRequired_notFiltering() {
     assertEquals(result, allPlaces);
   }
 
+  @Test
+  public void filter_webditeFilterNotRequired_notFiltering() {
```
szwbar: typo: website

project/src/test/java/com/google/sps/data/PlacesTest.java
szwbar: Missed that, but in any case I always prefer explicitness in tests :)

project/src/test/java/com/google/sps/data/PlacesTest.java
szwbar: Sorry, for some reason I thought two of them have the same status...

szwbar (approved):

project/src/test/java/com/google/sps/data/PlacesTest.java
karenav: fixed, thanks!

PR #46, opened 2020-11-24
Display places' Google Maps link

### LINKED ISSUE #43: Display Google URL to user if website isn't present

karenav:


project/src/main/webapp/script.js
```diff
@@ -134,28 +134,40 @@ function createPlaceElement(place) {
...
-    websiteLink.title = place.websiteUrl;
-    websiteLink.innerText = 'Restaurant\'s website';
-    placeElement.appendChild(websiteLink);
-  } else {
-    const noSite = document.createTextNode('We don\'t have a link to the restaurant\'s website.');
-    placeElement.appendChild(noSite);
+    addLinkToPlaceElement(placeElement, place.websiteUrl, 'Website');
+  }
+  if (place.googleUrl) {
+    addLinkToPlaceElement(placeElement, place.googleUrl, 'Google Maps link');
```
tbarnahor, line 142: I think 'Google Maps link' is a confusing description (since this doesn't lead to google maps). We can use the description of this field fromm the Places API: "Google page link"

project/src/main/webapp/script.js
```diff
@@ -134,28 +134,40 @@ function createPlaceElement(place) {
...
 }
 
+/**
+ * Returns a link that represents the given url and shows the given text.
+ */
+function addLinkToPlaceElement(placeElement, url, linkText) {
+  placeElement.appendChild(document.createElement('br'));
+  const link = document.createElement('a');
+  link.href = url;
+  link.title = url;
```
tbarnahor: We should check (when we deal with front end) that the urls aren't too long and ugly to be supplied as a title (displayed when hovering). This is especially for the google pages urls which tend to be long. I don't care that we leave it for now but just keep in mind this might won't look good and then we can change the title to a description (like "go to link") or just not supply it.

project/src/main/webapp/script.js
karenav: Interesting point, thought it does lead to Google Maps. Did you try to run this and press the link? I think this is a correct description

project/src/main/webapp/script.js
karenav: That's a good point. I don't mind not adding the title at all, or using a general term like "go to the website". Do you have a preference?

project/src/main/webapp/script.js
tbarnahor: Yeah it opens the map with the place's google page on the side. I now see the address is google maps so technically you are right, but I think when the user will see this description he won't understand this is to get more information about the place (such as opening hours, reviews and more info that appears on this link). I don't have strong feelings about it so we can leave like this if you prefer haha. And maybe one day we pilot and get actual users feedback about it :)

project/src/main/webapp/script.js
tbarnahor: I think we can give up on the title

tbarnahor (approved):

project/src/main/webapp/script.js
karenav: That's a great idea! You are right, maybe it's clear for me but not for everyone.
I'll leave it for now

project/src/main/webapp/script.js

szwbar (approved):

PR #47, opened 2020-11-30
Create Data accessor

### LINKED ISSUE #42: Add first version of data accessor (data getter + setter)

karenav:


project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @VisibleForTesting
+  DataAccessor(DatastoreService datastore) {
+    this.datastoreService = datastore;
+  }
+
+  /**
+  * @param userId the id of the user that we want to check the registration status about
+  * @return whether the user is registered to our system, by checking whether their id was
```
tbarnahor, line 51: I think "whether his id"

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,71 @@
...
+        new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, userIdKey);
+    Query query = new Query(userEntityName).setFilter(userIdFilter).setKeysOnly();
+    PreparedQuery results = datastoreService.prepare(query);
+    return results.asList(FetchOptions.Builder.withDefaults()).size() > 0;
+  }
+
+  /**
+  * Registers the user in our system, by adding an entity that represents them to datastore.
+  *
+  * @param userId the id of the user that we want to check the registration status about
```
tbarnahor: I think this description isn't what you meant

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,71 @@
...
+  */
+  public void registerUserId(String userId) {
+    if (Strings.isNullOrEmpty(userId)) {
```
tbarnahor: Maybe if an empty String is passed in the userId it should be handled in someway so the Authentication servlet will know the registration didn't work. In what situation can this be empty?

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,110 @@
...
+  public void registerUserId_validUser_registerUser() {
+    String userId = "12345";
+
+    dataAccessor.registerUserId(userId);
+    PreparedQuery results = createPreparedQueryByUserId(userId);
+
+    assertEquals(results.asList(FetchOptions.Builder.withDefaults()).size(), 1);
+  }
+
+  private PreparedQuery createPreparedQueryByUserId(String userId) {
```
tbarnahor: This function seems exactly like the isRegistered function (added the assertion that comes after this on the `registerUserId_validUser_registerUser` test).
So this makes me wonder if this "copy" of the function here is really needed, and if there is in general anything to do with the dependency of the registerUserId testing on the isRegistered tests.
I am just pointing this out, I am not really sure what is best to do in this situation.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class DataAccessorTest {
```
tbarnahor, line 43: I bet it was pretty challenging to learn how to test the DB, so good job for doing that! This is actually quite undestandable and looks good to me.

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: As we did in the design doc, we refer to the user as they, them (I'm not sure if you saw Bar's comment about that)

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: Absolutely correct, fixed it :)

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: That's a great question. 
The registration servlet won't return anything, like whether the registration was successful, the user was already registered, etc, because it's not required and this information won't be used.  And this happens because this information has no direct implications on the user or what we show him.
This can be empty if there was a problem with handling the user id in the functions that call this function (for example the token wasn't verified successfully). Your comment made me think that maybe we shouldn't return here, but throw an error. what do you think?

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: Right. I thought about it, and so far I didn't find a better way to test if the entity was added. 
What I said to myself (which made me feel better about this) is that these tests check the desired behaviour, and so if someone decides to change the functions' implementation, the functions still must work as we originally expected, even if the implementation will be different. I'm not sure if this helps haha. LMK if you think of an improvement or a different way to do that.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: Glad to hear that!

project/src/test/java/com/google/sps/data/DataAccessorTest.java
tbarnahor: That makes sense, and I don't have a better solution. I guess there is no way to avoid the dependency between `registerUserId` and `isRegistered` when testing.

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: Yes I think that an error should be thrown if the user tried to register and for some reason didn't succeed, but maybe in earlier stages and not here.

And a question since you brought it up - in the next milestone we will need to know out side of the registration Servlet if the user is registered in order to show him the old/new option, etc. I didn't find in the doc how it's done if there is no output from the servlet (we can also discuss this 1:1 when we get to it, just got a little confused).

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: Oh I didn't see Bar's comment.. I now see that we refer this way in parts of the design doc. 
To me it looks a bit wierd to move from single ("user") to plural ("their") in the same sentence, but what ever you guys think.

tbarnahor (approved):

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: Yeah, I guess it's the appropriate way to do that :)

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: Yes that's an interesting point, yet it's unrelated to this PR, we'll discuss it 1:1 :)
About the error - I'm wondering if it should be an invalid input (IllegalArgumentException) in **this function**. I think it should. I'm changing this. Please LMK if you think otherwise.

tbarnahor: Sounds good

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+  DataAccessor(DatastoreService datastore) {
+    this.datastoreService = datastore;
+  }
+
+  /**
+  * @param userId the id of the user that we want to check the registration status about
+  * @return whether the user is registered to our system, by checking whether their id was
+  *     previously added to datastore.
+  */
+  public boolean isRegisteredId(String userId) {
```
arielz-g: I'd go for just "isRegstered" (the ID is a parameter, so obvious in that sense.

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+    if (Strings.isNullOrEmpty(userId)) {
+      return false;
```
arielz-g: Isn't this an error? There should be a difference between "this user isn't registered" and "this is an invalid argument". You can use this library: https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html to statically import checkArgument which accepts a condition that must be true and a custom error message that follows.

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.repackaged.com.google.api.client.util.Strings;
+
+public class DataAccessor {
+
+  private final DatastoreService datastoreService;
+  @VisibleForTesting
+  final String userEntityName = "User";
```
arielz-g: Can you elaborate on this? Why is this the same kind for all requests? If it is, it should be a static final, no?

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+  */
+  public boolean isRegisteredId(String userId) {
+    if (Strings.isNullOrEmpty(userId)) {
+      return false;
+    }
+    Key userIdKey = KeyFactory.createKey(userEntityName, userId);
+    Filter userIdFilter =
+        new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, userIdKey);
+    Query query = new Query(userEntityName).setFilter(userIdFilter).setKeysOnly();
+    PreparedQuery results = datastoreService.prepare(query);
```
arielz-g: Suggestion: can be inline with the next line.

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+    return results.asList(FetchOptions.Builder.withDefaults()).size() > 0;
+  }
+
+  /**
+  * Registers the user in our system, by adding an entity that represents them to datastore.
+  *
+  * @param userId the id of the user that we want to register to our system
+  */
+  public void registerUserId(String userId) {
```
arielz-g: Same as previous method, can be registerUser.

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,74 @@
...
+    if (Strings.isNullOrEmpty(userId)) {
```
arielz-g: Use checkState (it'll throw the exception for you).

project/src/main/java/com/google/sps/data/DataAccessor.java
arielz-g: I'm missing a check here that this user isn't registered already. You shouldn't make assumptions regarding your input; think of it this way - anyone can call this method and which kind of situations do you wish to prevent? In your case it's rather simple as you already have a validating method in here. 

Note that there should be a fitting test for this as well.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,113 @@
...
+    String unRegisteredUserId = "54321";
+    Entity userEntity = new Entity(dataAccessor.userEntityName, registeredUserId);
+    datastoreService.put(userEntity);
+
+    assertFalse(dataAccessor.isRegisteredId(unRegisteredUserId));
+  }
+
+  @Test
+  public void isRegisteredId_invalidUserId_false() {
+    assertFalse(dataAccessor.isRegisteredId(""));
```
arielz-g: Adding to my previous comment, this should throw an exception, and conventionally I'd separate this to two tests (isRegistered_nullUserID_throws and isRegistered_emptyId_throws).

It makes little difference, but the way assertions work is that if one fails the test stops, so say you've got a failure for the empty string, the check for the null part isn't made (though it might go through as there were some code changes or what not). Simply put, it's easier to follow what went wrong.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Test
+  public void registerUserId_validUserId_registerUser() {
+    String userId = "12345";
+
+    dataAccessor.registerUserId(userId);
+    PreparedQuery results = createPreparedQueryByUserId(userId);
+
+    assertEquals(results.asList(FetchOptions.Builder.withDefaults()).size(), 1);
```
arielz-g: Can you also validate that the registered user has the userId you've set in the test?

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private PreparedQuery createPreparedQueryByUserId(String userId) {
```
arielz-g: Push this method to the end of the file (test functions should appear before helper test methods).

project/src/main/java/com/google/sps/data/DataAccessor.java
szwbar: It's the common way to refer to a person in a gender-neutral way. It's very accepted at Google, although it is surprising grammatically.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,113 @@
...
+    Key userIdKey = KeyFactory.createKey(dataAccessor.userEntityName, userId);
+    Filter userIdFilter =
+        new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, userIdKey);
+    Query query = new Query(dataAccessor.userEntityName).setFilter(userIdFilter).setKeysOnly();
+    return datastoreService.prepare(query);
+  }
+
+  @Test
+  public void registerUserId_invalidUserId_throwIllegalArgumentException() {
```
arielz-g: Same as before, may separate to two tests.

arielz-g (commented):
Nice testing!

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: That's a great idea, done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: It's the name of the entity as we want to keep the entity in datastore. I made it static, because as you said it's the same for all requests

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
arielz-g: Make it all caps and you're done (that's the convention for static finals).

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: That's a great point. Added.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: Very interesting, I didn't think about it. Seperated

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: Yes, done

project/src/test/java/com/google/sps/data/DataAccessorTest.java
karenav: Done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: Right, done!

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -0,0 +1,78 @@
...
+        .size() > 0;
+  }
+
+  /**
+  * Registers the user in our system, by adding an entity that represents them to datastore.
+  *
+  * @param userId the id of the user that we want to register to our system
+  */
+  public void registerUser(String userId) {
+    if (isRegistered(userId)) {
```
arielz-g: This needs to be in a checkArgument as well, think of it this way: it's invalid to call registerUser on an already registered user, you'd like an alert in this case, not to silence this as if everything went through fine.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertThrows(IllegalArgumentException.class, () -> dataAccessor.registerUser(""));
+  }
+
+  @Test
+  public void registerUser_nullUserId_throwIllegalArgumentException() {
+    assertThrows(IllegalArgumentException.class, () -> dataAccessor.registerUser(null));
+  }
+
+  @Test
+  public void registerUser_alreadyRegistered_dontRegisterAgain() {
```
arielz-g: Assert it throws per the comment above.

karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

arielz-g (approved):

PR #49, opened 2020-12-02
User verifier - use Google services to verify users' token and get their user Id 

karenav:
Not sure about the class' name, and so far I didn't find a more accurate one. Would love to hear comments about anything.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -72,7 +72,30 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>23.0</version>
+      <version>29.0-jre</version>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.api-client/google-api-client -->
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
+      <version>1.23.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava-jdk5</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.oauth-client/google-oauth-client-jetty -->
+    <dependency>
+      <groupId>com.google.oauth-client</groupId>
+      <artifactId>google-oauth-client-jetty</artifactId>
+      <version>1.23.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud</artifactId>
+      <version>0.10.0-alpha</version>
     </dependency>
   </dependencies>
 
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -64,7 +64,7 @@ public class UserVerifier {
    * the user's id, if the process was successful.
    *
    * @param idToken the token of the user
-   * @return an optional that holds the verified user ID, or null if the process wasn't successful
+   * @return an Optional with the verified user ID in case of success (and an empty optional else)
    */
   public Optional<String> getUserIdByToken(String idToken) {
     GoogleIdToken googleIdToken;
@@ -74,14 +74,14 @@ public class UserVerifier {
       return Optional.empty();
     }
     return googleIdToken != null
-        ? getSubjectFromPayload(googleIdToken.getPayload())
+        ? Optional.of(getSubjectFromPayload(googleIdToken.getPayload()))
         : Optional.empty();
   }
 
   @VisibleForTesting
   // This function wraps the call to GoogleIdToken.Payload.getSubject(). That's done in order to
   // enable testing, because the function is final and cannot be mocked.
-  Optional<String> getSubjectFromPayload(GoogleIdToken.Payload payload) {
-    return Optional.of(payload.getSubject());
+  String getSubjectFromPayload(GoogleIdToken.Payload payload) {
+    return payload.getSubject();
   }
 }
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -32,7 +32,8 @@ public class UserVerifier {
    * A private constructor that takes in a GoogleIdTokenVerifier object.
    * @param googleVerifier a Google tool for verifying the validity of Google users' tokens
    */
-  private UserVerifier(GoogleIdTokenVerifier googleVerifier) {
+  @VisibleForTesting
+  UserVerifier(GoogleIdTokenVerifier googleVerifier) {
     this.verifier = googleVerifier;
   }
 
@@ -47,16 +48,6 @@ public class UserVerifier {
             .setAudience(Collections.singletonList(clientId))
             .build()
     );
-  }
-
-  /** A constructor meant for enabling mockito to spy this class, so must have no parameters. */
-  @VisibleForTesting
-  UserVerifier() { }
-
-  /** Allows to inject the googleVerifier, meant for enabling mockito to spy and test this class. */
-  @VisibleForTesting
-  void updateGoogleVerifier(GoogleIdTokenVerifier googleVerifier) {
-    this.verifier = googleVerifier;
   }
 
   /**
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -28,6 +28,10 @@ public class UserVerifier {
 
   private GoogleIdTokenVerifier verifier;
 
+  /**
+   * A constructor that takes in a GoogleIdTokenVerifier object.
+   * @param googleVerifier a Google tool for verifying the validity of Google users' tokens
+   */
   @VisibleForTesting
   UserVerifier(GoogleIdTokenVerifier googleVerifier) {
     this.verifier = googleVerifier;
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -50,7 +50,7 @@ public class UserVerifier {
    * Autheticates the given token using Google's verifier, and returns an Optional that holds
    * the user's id, if the process was successful.
    *
-   * @param id_token the token of the user
+   * @param idToken the token of the user
    * @return an optional that holds the verified user ID, or null if the process wasn't successful
    */
   public Optional<String> getUserIdByToken(String idToken) {
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -29,11 +29,10 @@ public class UserVerifier {
   private GoogleIdTokenVerifier verifier;
 
   /**
-   * A constructor that takes in a GoogleIdTokenVerifier object.
+   * A private constructor that takes in a GoogleIdTokenVerifier object.
    * @param googleVerifier a Google tool for verifying the validity of Google users' tokens
    */
-  @VisibleForTesting
-  UserVerifier(GoogleIdTokenVerifier googleVerifier) {
+  private UserVerifier(GoogleIdTokenVerifier googleVerifier) {
     this.verifier = googleVerifier;
   }
 
@@ -48,6 +47,16 @@ public class UserVerifier {
             .setAudience(Collections.singletonList(clientId))
             .build()
     );
+  }
+
+  /** A constructor meant for enabling mockito to spy this class, so must have no parameters. */
+  @VisibleForTesting
+  UserVerifier() { }
+
+  /** Allows to inject the googleVerifier, meant for enabling mockito to spy and test this class. */
+  @VisibleForTesting
+  void updateGoogleVerifier(GoogleIdTokenVerifier googleVerifier) {
+    this.verifier = googleVerifier;
   }
 
   /**
@@ -65,7 +74,14 @@ public class UserVerifier {
       return Optional.empty();
     }
     return googleIdToken != null
-        ? Optional.of(googleIdToken.getPayload().getSubject())
+        ? getSubjectFromPayload(googleIdToken.getPayload())
         : Optional.empty();
+  }
+
+  @VisibleForTesting
+  // This function wraps the call to GoogleIdToken.Payload.getSubject(). That's done in order to
+  // enable testing, because the function is final and cannot be mocked.
+  Optional<String> getSubjectFromPayload(GoogleIdToken.Payload payload) {
+    return Optional.of(payload.getSubject());
   }
 }
--- a/project/src/main/java/com/google/sps/data/UserVerifier.java
+++ b/project/src/main/java/com/google/sps/data/UserVerifier.java
@@ -17,11 +17,11 @@ package com.google.sps.data;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.util.Collections;
+import java.util.Optional;
 import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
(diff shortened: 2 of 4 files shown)


project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+
+public class UserVerifierTest {
+
+  private GoogleIdTokenVerifier googleVerifier = mock(GoogleIdTokenVerifier.class);
```
tbarnahor: Shouldn't this be a constant? (static final, uppercase)

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.api.client.json.jackson.JacksonFactory;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.api.client.http.javanet.NetHttpTransport;
+
+public class UserVerifier {
+
+  private GoogleIdTokenVerifier verifier;
+
+  @VisibleForTesting
+  UserVerifier(GoogleIdTokenVerifier googleVerifier) {
```
tbarnahor, line 36: I think this should be documented (or at least the verifier field on line 29) . What is this verifier?

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Autheticates the given token using Google's verifier, and returns an Optional that holds
+   * the user's id, if the process was successful.
+   *
+   * @param idToken the token of the user
+   * @return an optional that holds the verified user ID, or null if the process wasn't successful
+   */
+  public Optional<String> getUserIdByToken(String idToken) {
```
tbarnahor, line 60: Nice idea to use an Optional

tbarnahor (approved):
Nice job! And I think the name is fine

project/src/test/java/com/google/sps/data/UserVerifierTest.java
karenav: good point, fixed!

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: thanks!

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: added :)

project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,74 @@
...
+  public void getUserIdByToken_emptyIdToken_emptyOptional() {
+    UserVerifier userVerifier = createUserVerifierWithGoogleVerfier();
+    Optional<String> result = userVerifier.getUserIdByToken("");
+
+    assertFalse(result.isPresent());
+  }
+
+  @Test
+  public void getUserIdByToken_nullIdToken_emptyOptional() {
+    UserVerifier userVerifier = createUserVerifierWithGoogleVerfier();
```
tbarnahor: I think this can be created once and used for all tests (it can also be spied on instead of spying the class).

karenav: I thought it would be less readable, but if you're commenting about it I guess it's not doing what I hoped, so np :)

project/pom.xml
```diff
@@ -72,7 +72,30 @@
...
       <artifactId>guava</artifactId>
-      <version>23.0</version>
+      <version>29.0-jre</version>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.api-client/google-api-client -->
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
+      <version>1.23.0</version>
+      <exclusions>
```
szwbar, line 82: Did you exclude Guava since you already have a dependency on it?

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,87 @@
...
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+import java.util.Optional;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdTokenVerifier;
+import com.google.api.client.json.jackson.JacksonFactory;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.api.client.http.javanet.NetHttpTransport;
+
+public class UserVerifier {
```
szwbar, line 27: Naming suggestion: UserAuthenticator? 

The current name is OK but you asked for suggestions :)

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,87 @@
...
+  void updateGoogleVerifier(GoogleIdTokenVerifier googleVerifier) {
+    this.verifier = googleVerifier;
+  }
+
+  /**
+   * Autheticates the given token using Google's verifier, and returns an Optional that holds
+   * the user's id, if the process was successful.
+   *
+   * @param idToken the token of the user
+   * @return an optional that holds the verified user ID, or null if the process wasn't successful
```
szwbar: When you say "null" you mean "empty optional"?

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return googleIdToken != null
+        ? getSubjectFromPayload(googleIdToken.getPayload())
+        : Optional.empty();
+  }
+
+  @VisibleForTesting
+  // This function wraps the call to GoogleIdToken.Payload.getSubject(). That's done in order to
+  // enable testing, because the function is final and cannot be mocked.
+  Optional<String> getSubjectFromPayload(GoogleIdToken.Payload payload) {
```
szwbar: This function always returns a `String`. I suggest returning `String` here and wrapping it with an `Optional` in `getUserIdByToken`.

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,87 @@
...
+      return Optional.empty();
+    }
+    return googleIdToken != null
+        ? getSubjectFromPayload(googleIdToken.getPayload())
+        : Optional.empty();
+  }
+
+  @VisibleForTesting
+  // This function wraps the call to GoogleIdToken.Payload.getSubject(). That's done in order to
+  // enable testing, because the function is final and cannot be mocked.
```
szwbar, line 74: I don't really understand why you need this distinction. 
AFAIU, `Payload` has a `setSubject` method that simply changes the subject. Is that right? You can make the verifier return the payload you're interested in, and set its subject by calling `setSubject`. 

Please LMK if I'm missing something.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -38,6 +38,8 @@
 @WebServlet("/query")
 public final class QueryServlet extends HttpServlet {
 
+  private static final long serialVersionUID = 1L;
```
szwbar, line 41: I don't really think you need that...

project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+
+public class UserVerifierTest {
+
+  private static final GoogleIdTokenVerifier GOOGLE_VERIFIER = mock(GoogleIdTokenVerifier.class);
```
szwbar: Are you 100% sure that `GoogleIdTokenVerifier` is immutable once it is declared `final`? 
Usually the naming convention for non-immutable member fields is camelCase, and use all caps snake case when the field is strictly immutable (e.g. `ImmutableMap` or `String`).

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /** A constructor meant for enabling mockito to spy this class, so must have no parameters. */
+  @VisibleForTesting
+  UserVerifier() { }
+
+  /** Allows to inject the googleVerifier, meant for enabling mockito to spy and test this class. */
+  @VisibleForTesting
+  void updateGoogleVerifier(GoogleIdTokenVerifier googleVerifier) {
+    this.verifier = googleVerifier;
+  }
```
szwbar: Why not merge the constructor and `updateGoogleVerifier`? Simply make the constructor accept a `GoogleIdTokenVerifier`.

project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final UserVerifier USER_VERIFIER = new UserVerifier();
+
+  @Before
+  public void setUp() {
+    USER_VERIFIER.updateGoogleVerifier(GOOGLE_VERIFIER);
+  }
+
+  @Test
+  public void getUserIdByToken_emptyIdToken_emptyOptional() {
```
szwbar: Naming nit: `..._returnsEmptyOptional`

project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+    Optional<String> result = USER_VERIFIER.getUserIdByToken(null);
+
+    assertFalse(result.isPresent());
+  }
+
+  @Test
+  public void getUserIdByToken_validIdToken_getUserId() throws Exception {
+    UserVerifier spiedUserVerifier = spy(USER_VERIFIER);
+    GoogleIdToken mockedToken = mock(GoogleIdToken.class);
+    Payload mockedPayload = mock(Payload.class);
```
szwbar: I think you are over-mocking things. 

You can initialize a `GoogleIdToken` class yourself, with the `Payload` object you expect to have. 
Then, the only mock you have is the `GOOGLE_VERIFIER`, you don't need to spy on the tested class, and you end up with as-close-to-real implementation as you can.

szwbar (changes requested):
Generally looks good! 
I'm not sure about the pattern you chose in the test, feel free to talk F2F about it.

project/pom.xml
karenav: Yes. The double dependency caused issues and made other parts of the code to stop working.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
karenav: There is an annoying warning that appears as long as we don't have that haha. I know it's not needed, that's why it wasn't there so far. It's just annoying to work with, and I think it's harmless to add this.

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: Because then we can't spy the class, it demands a default constructor

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: That's an interesting idea. When we do that, we expect something from the behaviour of the Payload class: we expect that if we set the subject, then that's exactly the subject we'll get afterwards (and that's an expectation that we don't use in our code, only in the tests). And I'm not sure how much I want to rely on that, rather than specify exactly the outcome that I want to see. WDYT?

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: Yes, I updated the documentation to be clearer

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: done

project/src/test/java/com/google/sps/data/UserVerifierTest.java
karenav: done

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: I took up on you offer, updated.

project/src/test/java/com/google/sps/data/UserVerifierTest.java
karenav: The structure changed due to other comments, no longer relevant :)

karenav: The structure changed a bit, is this still relevant?

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: done

project/src/test/java/com/google/sps/data/UserVerifierTest.java
karenav: changed. LMK if you think it still needs work

project/src/main/java/com/google/sps/data/UserVerifier.java
karenav: Yes, that's the alternative I considered, but I feel verify is slightly more accurate here than authenticate. Thank you!

project/pom.xml
szwbar: Thanks for explaining!

project/src/main/java/com/google/sps/servlets/QueryServlet.java
szwbar: Ack.

project/src/test/java/com/google/sps/data/UserVerifierTest.java
szwbar: Yes, as it concerns the naming of the variable, which IMO should be camelCase.

project/src/main/java/com/google/sps/data/UserVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.api.client.json.jackson.JacksonFactory;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.api.client.http.javanet.NetHttpTransport;
+
+public class UserVerifier {
+
+  private GoogleIdTokenVerifier verifier;
+
+  /**
+   * A private constructor that takes in a GoogleIdTokenVerifier object.
```
szwbar: nit: test-only constructor (not `private`).

project/src/test/java/com/google/sps/data/UserVerifierTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+
+public class UserVerifierTest {
+
+  private static final GoogleIdTokenVerifier MOCKED_GOOGLE_VERIFIER
```
szwbar: Following up on a comment: the name should be camelCase and not SNAKE_CASE.

project/src/main/java/com/google/sps/data/UserVerifier.java
szwbar: Great so it worked! IMO is a better test now. 

I don't think you need this method anymore (`getSubjectFromPayload`) and I suggest merging it with the existing code.

szwbar (approved):
Great work! Thanks for following up on my testing suggestions. 
Just a few more nit.

PR #51, opened 2020-12-06
Google Sign in button

### LINKED ISSUE #41: Add Google sign in and register to site

karenav:
Important - this shouldn't be merged before the registration servlet is merged, because an error will be thrown if a registration servlet won't be found.

### DIFF
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -16,7 +16,13 @@ package com.google.sps.data;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
@@ -26,13 +32,13 @@ import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.repackaged.com.google.api.client.util.Strings;
 
 public class DataAccessor {
 
   private final DatastoreService datastoreService;
   @VisibleForTesting
   static final String USER_ENTITY_NAME = "User";
+  static final String RECOMMENDATION_ENTITY_NAME = "Recommendation";
 
   /**
    * A constructor that creates a DatastoreService instance for the class.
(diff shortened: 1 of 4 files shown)


project/src/main/webapp/index.html
```diff
@@ -1,17 +1,25 @@
...
+    <meta charset="UTF-8">
+    <title>food-mood</title>
+    <link rel="stylesheet" href="style.css">
+    <base target="_blank" rel="noopener noreferrer">
+  </head>
+  <body class="text-content">
+    <h1> FoodMood - your food order helper!</h1>
+    <div id="sign-in-segment">
+      <div id="sign-in-button" class="g-signin2" data-onsuccess="onSignIn" data-prompt="select_account"></div>
+      <div id ="user-welcome-messeage-container"> Hello! You are welcome to sign in </div>
```
tbarnahor: message

project/src/main/webapp/script.js
```diff
@@ -360,3 +369,29 @@ function addPlaceMarker(map, place) {
...
+function signOut() {
+  gapi.auth2.getAuthInstance().signOut();
+  document.getElementById('user-welcome-messeage-container').innerText =
+      'You are currently not logged in with a Google account.';
+  googleUser = null;
+}
+
+/** Registers the logged in user, using the registration servlet. */
+function registerUserByToken() {
+  if (!googleUser) {
```
tbarnahor, line 393: If this is already checked before this method is called so maybe this isn't needed.

karenav: You are right, it's checked twice. I believe in double checking to avoid errors. Do you think it's better to have only one?

project/src/main/webapp/index.html
karenav: right, thanks! fixed

project/src/main/webapp/script.js
tbarnahor: I would probably use just one since the same is checked right before and right after entering the function, but if you like it it's fine :)

tbarnahor (approved):

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -47,26 +48,26 @@ public DataAccessor() {
   }
 
   /**
-  * @param userId the id of the user that we want to check the registration status about
-  * @return whether the user is registered to our system, by checking whether their id was
-  *     previously added to datastore.
-  */
+   * @param userId the id of the user that we want to check the registration
```
szwbar: nit: In comments, write "ID"

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -47,26 +48,26 @@ public DataAccessor() {
...
+   *               status about
```
szwbar: nit: about -> for

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -26,13 +27,13 @@
...
 import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.repackaged.com.google.api.client.util.Strings;
 
 public class DataAccessor {
 
   private final DatastoreService datastoreService;
   @VisibleForTesting
   static final String USER_ENTITY_NAME = "User";
+  static final String RECOMMENDATION_ENTITY_NAME = "Recommendation";
```
szwbar: You don't seem to use this String.

project/src/main/webapp/index.html
```diff
@@ -1,17 +1,25 @@
...
+    <title>food-mood</title>
+    <link rel="stylesheet" href="style.css">
+    <base target="_blank" rel="noopener noreferrer">
+  </head>
+  <body class="text-content">
+    <h1> FoodMood - your food order helper!</h1>
+    <div id="sign-in-segment">
+      <div id="sign-in-button" class="g-signin2" data-onsuccess="onSignIn" data-prompt="select_account"></div>
+      <div id ="user-welcome-message-container"> Hello! You are welcome to sign in </div>
+      <button id="sign-out-button" onclick="signOut();">Sign out</button>
```
szwbar: Is the sign out button always displayed? Shouldn't you show it only if the user is signed out?

project/src/main/webapp/script.js
```diff
@@ -38,6 +44,9 @@ function fetchFromQuery() {
   displayResultsPage();
   const userCoords = JSON.parse(localStorage.getItem('userLocation'));
   const map = createMap({lat: userCoords.lat, lng: userCoords.lng});
+  if (googleUser) {
+    registerUserByToken();
+  }
```
szwbar: Why does the registration phase happens when fetching from the server? 
I think it should happen directly after the signing is successful from Google. WDYT?

szwbar (changes requested):

project/src/main/webapp/script.js
karenav: I don't really mind, done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: (These are remains of a different PR, I restored the file to the version that's in the master)

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: (These are remains of a different PR, I restored the file to the version that's in the master)

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: (These are remains of a different PR, I restored the file to the version that's in the master)

project/src/main/webapp/index.html
karenav: Great idea! done

project/src/main/java/com/google/sps/data/DataAccessor.java
szwbar: I still see "id" and not "ID" though.

project/src/main/webapp/script.js
szwbar: I think it has implications on multiple things in the flow, such as UI responsiveness and latency for the user.

szwbar (approved):

PR #52, opened 2020-12-07
Warning that Mac users might have an inaccurate location

karenav:


tbarnahor (approved):

szwbar (approved):
LGTM, although I think you can use the User Agent to identify yourself if the user is using a Mac! 
I recommend reading about it.

PR #53, opened 2020-12-07
Filter by rounding rating to closest int

karenav:


project/src/test/java/com/google/sps/data/PlacesTest.java
```diff
@@ -89,6 +89,29 @@ public void filter_tooLowRating_filterOut() {
     assertEquals(result, ImmutableList.of(highRatingPlace));
   }
 
+  @Test
+  public void filter_floatingPointRating_filterOutByRoundingToNearestInt() {
+    // This test checks that the rating is filtered after rounding it to the nearest int.
+    // For example, when places with 5 starts are desired, places that have at least 4.5 stars will
```
tbarnahor: There is a little issue with rint (that I am fine with, just pointing it out) that in case of 0.5 at decimal places it will round off to nearest even double value. This means that 4.5 will round to 4 and will not be supplied if 5 was the threshold.

karenav: Yes you are right, I also noticed it. I changed the documentation to make it clearer. Other than that do you think something should be done?

tbarnahor: No I think its fine, just the documentation didn't exactly fit but now it's good :)

tbarnahor (approved):

project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -58,7 +58,7 @@
         places.stream()
             .sorted(Comparator.comparing(Place::rating).reversed())
             .filter(place -> place.businessStatus() == BusinessStatus.OPERATIONAL)
-            .filter(place -> place.rating() >= minRating)
+            .filter(place -> Math.rint(place.rating()) >= minRating)
```
szwbar: So for a place with rating 3.8, the Math.rint result would be 4. 
Then, if `minRating == 4`, then this evaluates to `true`, although 3.8 < 4. 
Is this the intended behavior?

karenav: Yes, that's the wanted behaviour (that's the purpose of the PR).
So for example, if someone specifies they want places with 5 starts, they'll get results (otherwise they probably won't get anything, because almost no place actually has 5 stars).

szwbar: Right, so I think in that case that `minRating` is a confusing name. Maybe rename it? 
Something like `approxMinRating` or similar.

karenav: good point. done

project/src/main/webapp/index.html
```diff
@@ -45,7 +45,7 @@ <h1> FoodMood - your food order helper!</h1>
         <label for="rating-5">5 star</label>
       </form>
       <form id="price-form" class="form-element-toolbar">
-        <p>Maximal price level:</p>
+        <p>Desired price level:</p>
```
szwbar, line 48: I liked the "desired" terminology!

szwbar (approved):

PR #54, opened 2020-12-07
Registration servlet

karenav:


project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.Optional;
+import com.google.sps.data.DataAccessor;
+import com.google.sps.data.UserVerifier;
+import com.google.common.annotations.VisibleForTesting;
+
+ /**
```
tbarnahor: Something wierd with indentations. 
Also: "A servlet that registers.."

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -0,0 +1,64 @@
...
+@WebServlet("/register")
+public final class RegistrationServlet extends HttpServlet {
+
+  private static final long serialVersionUID = 1L;
+  private UserVerifier userVerifier;
+  private DataAccessor dataAccessor;
+
+  @Override
+  public void init() {
+    this.userVerifier = UserVerifier.create("CLIENT_ID");
```
tbarnahor: Why is this constant always passed? Is this "A Google API client ID"?

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -0,0 +1,64 @@
...
+  void init(UserVerifier userVerifier, DataAccessor dataAccessor) {
+    this.userVerifier = userVerifier;
+    this.dataAccessor = dataAccessor;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userIdToken = request.getParameter("idToken");
+    if (userIdToken == null) {
+      return;
```
tbarnahor, line 53: If the IdToken passed to the getUserIdByToken is never null, I think the null test in the verifier might be redundent (or, this check is redundent since this null situation is handeled in the verifier).

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.data.DataAccessor;
+import com.google.sps.data.UserVerifier;
+
+@RunWith(JUnit4.class)
+public class RegistrationServletTest {
+
+  private static final HttpServletRequest REQUEST = mock(HttpServletRequest.class);
+  private static final HttpServletResponse RESPONSE = mock(HttpServletResponse.class);
+  private UserVerifier userVerifier = mock(UserVerifier.class);
```
tbarnahor: I think those should be constants

tbarnahor (commented):
Nice tests! Very clear

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: We want to make this code bullet-proof to any changes someone might do with our code. This makes sure that even if someone changes the way the servlet is called, or calls the servlet from a new place in the code, it won't cause errors (in this case - NullPointerException). 
(Not to mention that this is a scenario that we want to test, so the function must behave this way...)
In other words - even if something is currently avoided, we still have to make sure we handle it.

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
karenav: Good point, changed

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: It's like an environment variable (that holds our private credentials), but so far we haven't managed to really use it that way, so it is changed manually when we run the app.

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: Thanks! fixed

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
tbarnahor: Ohhh Ok I didn't understand that.

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
tbarnahor: Yes I agree that it's good that the verifier handels getting null, just thought that since the verifier handels the null case then this first if statement isn't needed, but I understand you want that double check.

tbarnahor (approved):

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+import com.google.sps.data.DataAccessor;
+import com.google.sps.data.UserVerifier;
+import com.google.common.annotations.VisibleForTesting;
+
+/** A servlet that registers a user, according to a given token. */
+@WebServlet("/register")
+public final class RegistrationServlet extends HttpServlet {
+
+  private static final long serialVersionUID = 1L;
```
szwbar: Do you need this?

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
szwbar: Maybe add an error code to the response, marking that you got an invalid argument?

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userIdToken = request.getParameter("idToken");
+    if (userIdToken == null) {
+      return;
+    }
+    Optional<String> optionalUserId = userVerifier.getUserIdByToken(userIdToken);
+    if (!optionalUserId.isPresent()) {
+      return;
```
szwbar, line 59: Error message?

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.sps.data.DataAccessor;
+import com.google.sps.data.UserVerifier;
+
+@RunWith(JUnit4.class)
+public class RegistrationServletTest {
+
+  private static final HttpServletRequest REQUEST = mock(HttpServletRequest.class);
+  private static final HttpServletResponse RESPONSE = mock(HttpServletResponse.class);
+  private static final UserVerifier USER_VERIFIER = mock(UserVerifier.class);
+  private static final DataAccessor DATA_ACCESSOR = mock(DataAccessor.class);
```
szwbar: Naming: userVerifier and dataAccessor. 

Also, it's customary to prefix with `mock`: `mockUserVerifier`, etc.

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+  private RegistrationServlet servlet;
+
+  @Before
+  public void setUp() throws Exception {
+    servlet = new RegistrationServlet();
+    servlet.init(USER_VERIFIER, DATA_ACCESSOR);
+  }
+
+  @Test
+  public void doPost_validToken_registerUser() throws Exception {
```
szwbar: naming nit `registersUser`

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+    String idToken = "abcde";
+    String userId = "12345";
```
szwbar, line 55: Make those class-level constants.

szwbar (changes requested):
Looks good!

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
karenav: done

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
karenav: done

project/src/test/java/com/google/sps/servlets/RegistrationServletTest.java
karenav: done

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: added

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: found another way to ignore the warning

project/src/main/java/com/google/sps/servlets/RegistrationServlet.java
karenav: great idea, done

szwbar (approved):

PR #56, opened 2020-12-07
Durations scorer

tbarnahor:
This PR adds the calculation of driving duration to the scorer


project/src/main/java/com/google/sps/data/Places.java
```diff
@@ -46,7 +49,8 @@
 
   /**
    * Sorts the given list of Places by score.
-   * @param places the list we want to sort.
+   *
+   * @param places       the list we want to sort.
```
karenav: extra spaces

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -1,27 +1,54 @@
...
-            ImmutableList<Place> places, LatLng userLocation) {
-        ImmutableMap<Place, Double> durations = getDurations(places, userLocation);
+    public ImmutableMap<Place, Double> getScores(ImmutableList<Place> places, LatLng userLocation) {
+        ImmutableMap<Place, Long> durations;
+        try {
+            durations = getDurations(places, userLocation);
+        } catch (ApiException | InterruptedException | IOException e) {
+            // TODO(Tal): log error
+			return scoreByRating(places);
+		}
```
karenav: I think there's too much indentation

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -30,20 +57,52 @@
...
+        DistanceMatrixApiRequest distanceRequest =
+            DistanceMatrixApi.newRequest(context)
+                .origins(origins)
+                .destinations(destination)
+                .mode(TravelMode.DRIVING);
+        DistanceMatrix distanceMatrix = getDistanceResults(distanceRequest);
+        for (int i = 0; i < places.size(); i++) {
+            durations.put(places.get(i), distanceMatrix.rows[i].elements[0].duration.inSeconds);
+        }
+        return ImmutableMap.copyOf(durations);
```
karenav: You can use the [immutable map builder](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html) instead of converting, I think it's a better practice :)

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -30,20 +57,52 @@
...
+
+
+    /**
+    * Queries Google Places API according to given query.
+    *
+    * @param distanceMatRequest A DistanceMatrixApiRequest with all places as origins
+    *     and the user's location as the destination
+    * @return A DistanceMatrix containig the distance and duration from each origin
+    *     to the destination, each row in the matrix corresponds to an origin
+    * @throws IOException
```
karenav: I think usually we elaborate what makes each exception be thrown

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -41,6 +41,7 @@
 import com.google.maps.model.LatLng;
 import com.google.sps.data.BusinessStatus;
 
+
```
karenav: extra space

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
```diff
@@ -43,16 +53,49 @@
...
     private static final ImmutableList<Place> PLACES_TO_SCORE = ImmutableList.of(PLACE_1, PLACE_2);
+    private static final DistanceMatrixRow[] DISTANCE_MATRIX_ROW =
+        createDistanceMatrixRows(2, 1800L); // 30 minutes durations
+
+    // Each DistanceMatrixRow represents the results for a single origin.
+    // Since there is only one destination (the user's location) for all origins (the places),
+    // each row will include one DistanceMatrixElement that holds the duration between a place
+    // to the user location.
+    // This method creates uniform durations between each origin to the destinations.
+    private static DistanceMatrixRow[] createDistanceMatrixRows(
```
karenav: I think usually helper functions go to the bottom of the file (after the tests)

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
```diff
@@ -43,16 +53,49 @@
     private static final LatLng USER_LOCATION = new LatLng(33.12, 34.56);
     private static final Place PLACE_1 = PLACE_BUILDER.setName("name1").setRating(3).build();
     private static final Place PLACE_2 = PLACE_BUILDER.setName("name2").setRating(5).build();
+    private static final String[] PLACES_ADDRESSES = {"Place1 Address", "Place2 Address"};
+    private static final String[] USERS_ADDRESS = {"User's Address"};
     private static final ImmutableList<Place> PLACES_TO_SCORE = ImmutableList.of(PLACE_1, PLACE_2);
+    private static final DistanceMatrixRow[] DISTANCE_MATRIX_ROW =
+        createDistanceMatrixRows(2, 1800L); // 30 minutes durations
```
karenav: If I understand correctly, and this 2 represents the size of PLACES_ADDRESSES, I would use PLACES_ADDRESSES.length, because I think it's more modular and readable

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
karenav: I think the name of the function can be a bit more informative, for example createUniformDistanceMatrixRows or something else that states that something generic is being created

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
```diff
@@ -61,10 +104,34 @@ public void getScores_validPlaceList_returnsMapOfCorrectScores() {
...
+        PlacesScorerImpl spiedScorer = spy(placesScorer);
+        doThrow(new IOException())
+            .when(spiedScorer)
+            .getDistanceResults(any(DistanceMatrixApiRequest.class));
+
+        ImmutableMap<Place, Double> result =
+            spiedScorer.getScores(PLACES_TO_SCORE, USER_LOCATION);
+
+        Double expectedScorePlace1 = 0.6;
+        Double expectedScorePlace2 = 1.0;
```
karenav: I guess this relies on the fact that you know that the ratings that you used are 3 and 5. I think this is unmodular, because if someone changes one of the ratings, the test will falsely fail. I think you should have variables that hold 3 and 5, and use these variables both when building the places and here (in a calculation).

karenav (commented):
I think it looks great!

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: I changed the places names to state their rating so maybe the expected result is clearer this way. I do think that it is better to calculate speratly and provide the expected result without doing the calculation on the test. I think it cleaner, and if the result changes from the calculated expected result, it is fine that the test will fail.

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: really good idea

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: true

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: I am not sure, I stuck to the structure of the PlacesFetcherTest where organizing all the constants used for testing is at the beggining.

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
tbarnahor: In the fetcher I only elaborated when it was our Fetcher exception, but I guess you are right.

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
tbarnahor: Totaly!

project/src/main/java/com/google/sps/data/Places.java
tbarnahor: Right, weird

karenav (approved):

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
```diff
@@ -37,34 +47,93 @@
...
-        .setPlaceId("placeid1")
+        .setPlaceId("placeid")
+        .setName("name")
         .setBusinessStatus(BusinessStatus.OPERATIONAL);
 
     private static final LatLng USER_LOCATION = new LatLng(33.12, 34.56);
-    private static final Place PLACE_1 = PLACE_BUILDER.setName("name1").setRating(3).build();
-    private static final Place PLACE_2 = PLACE_BUILDER.setName("name2").setRating(5).build();
-    private static final ImmutableList<Place> PLACES_TO_SCORE = ImmutableList.of(PLACE_1, PLACE_2);
+    private static final Place PLACE_RATING_3 = PLACE_BUILDER.setRating(3).build();
```
szwbar: I think it would be more readable if you create these places on-demand. 

It's useful to have place_1 and place_2 if these places are arbitrary; But if you have specific logic, sometimes it's better to create the objects in the test so it would be easier for the reader to understand the flow.

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
```diff
@@ -37,34 +47,93 @@
...
+    private static final Place PLACE_RATING_5 = PLACE_BUILDER.setRating(5).build();
+    private static final String[] PLACES_ADDRESSES = {"Place1 Address", "Place2 Address"};
```
szwbar, line 56: Why array and not ImmutableList?

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
```diff
@@ -30,20 +59,51 @@
...
+        LatLng[] origins = places.stream()
+            .map(place -> place.location()).toArray(LatLng[]::new);
+        DistanceMatrixApiRequest distanceRequest =
+            DistanceMatrixApi.newRequest(context)
+                .origins(origins)
+                .destinations(destination)
+                .mode(TravelMode.DRIVING);
+        DistanceMatrix distanceMatrix = getDistanceResults(distanceRequest);
+        for (int i = 0; i < places.size(); i++) {
+            durations.put(places.get(i), distanceMatrix.rows[i].elements[0].duration.inSeconds);
```
szwbar: Does the API guarantee that the distance matrix will contain all the elements? And that they're not null?

szwbar (commented):
Looks really good! Just a few comments.

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: Because it is sent to DistanceMatrix constructor, which gets arrays and not lists.

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
tbarnahor: That's a great point.
Rows length will always match the number of origin elements, regardless of whether they're routable, and so is the elements array. 
What happens in case of a problem in finding a route between a place and destination, is that the element will have a status that is different then "OK".
I added this check (and a a test), and for now a problematic place gets the max duration. I think this place should be filtered out, but I added this as a TODO for now because I think there are more urgent things to deal with.

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
tbarnahor: I agree, it is more readable now

project/src/test/java/com/google/sps/data/PlacesScorerImplTest.java
szwbar: Ack, thanks.

project/src/main/java/com/google/sps/data/PlacesScorerImpl.java
szwbar: 👍

szwbar (approved):
Great work!

PR #58, opened 2020-12-15
User feedback class

karenav:


project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /** @return a list of IDs of the places that were recommended to the user. */
+  public abstract ImmutableList<String> placesRecommendedToUser();
+
+  /** @return the ID of the place the user chose, if any. */
+  public abstract Optional<String> placeUserChose();
+
+  /**
+  * @return true if the user requested to try again and get more recommendations (implying they
+  * weren't satisfied with the recommenrations he recieved).
```
tbarnahor: recommendations

tbarnahor: And also refering user as a "he" (although IMO it's fine hahaha)

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -16,38 +16,68 @@
...
+  /**
+   * @param priceLevel the price level of a place, should be represented by a number between 0-4.
+   * @throws IllegalArgumentException
+   */
+  public static void validatePriceLevel(int priceLevel) throws IllegalArgumentException {
+    checkArgument(priceLevel >= MIN_PRICE_LEVEL && priceLevel <= MAX_PRICE_LEVEL,
         "Price level should be between %s-%s", MIN_PRICE_LEVEL, MAX_PRICE_LEVEL);
-    }
+  }
+
```
tbarnahor, line 53: extra space

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+     * Concludes the building of a new UserFeedback instance.
+     *
+     * @return the new instance.
+     * @throws IllegalArgumentException if the user feedback isn't consistent.
+     */
+    public UserFeedback build() throws IllegalArgumentException {
+      UserFeedback feedback = autoBuild();
+      ValidationUtils.validateChosenPlaceInReccomendedPlaces(feedback.placeUserChose(),
+          feedback.placesRecommendedToUser());
+      ValidationUtils.validateUserTriedAgainOnlyIfdidntchoose(feedback.placeUserChose(),
```
tbarnahor: This can actually easiliy happen, our UI doesn't restrict selecting a place and then pressing try again. If it really bothers us I think we should do something about it in the UI and not just get here errors. Options I thought about:
* somehow end the User's session after he picked a place (maybe add some "hope you enjoy" message and exit the results screen).
* When the user selects a place and then pressess "try again" inform him he can't try again while a place is selected

Another option, that I think is not so bad is to just treat this situation as the user did select that place, even if he pressed try again after wards. 

What do you think?

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,83 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import com.google.common.collect.ImmutableList;
```
tbarnahor: move to non static imports

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.junit.Assert.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Optional;
```
tbarnahor, line 22: move to non static imports

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private static final String PLACE_ID_1 = "12345";
+  private static final ImmutableList<String> PLACES_WITH_PLACE_ID_1 =
+      ImmutableList.of(PLACE_ID_1, "11111", "22222");
+
+  @Test
+  public void build_invalidBothChoosePlaceAndTryagain_throwsIllegalArgumentException() {
+    assertThrows(IllegalArgumentException.class, () -> {
+      UserFeedback.builder()
+          .setPlacesRecommendedToUser(ImmutableList.of(PLACE_ID_1, "11111", "22222"))
```
tbarnahor: You can use here `PLACES_WITH_PLACE_ID_1`

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: I think it's right to point this out. I'm actually working on the UI in a different PR (on the JS and HTML). I think currently I'm implementing the last option you said, but this can be changed. WDYT should be the way to choose? I think either the first or the last. 
Either way, I think the treatment isn't required for this current PR, I added a TODO to remind us this should happen

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
karenav: absolutely right, done

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
karenav: absolutely right, done

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
karenav: Right, thanks!

project/src/main/java/com/google/sps/data/ValidationUtils.java
karenav: right, thanks!

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: right! thanks

project/src/main/java/com/google/sps/data/UserFeedback.java
tbarnahor: The treatment that is required for this PR is if we choose the last option I mentioned (remove the validation).

I thought of another option which I think can be the best user experience but I don't think we have the time for this change:
When the user picks a place and then presses try again, then leave the option he chose and only replace the two other options. 

Since we are limited in time I would go for the last option, which all it takes is removing this validation, but if you are not comftable with this you can leave it as a todo and we discuss it later.

tbarnahor (approved):

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.Optional;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Represents the feedback that users enter about the recommendations they were given.
```
arielz-g: nit: how about "Represents the input feedback a user provides for a Food-Mood recommendation."

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+@AutoValue
+public abstract class UserFeedback {
+
+  /** @return a list of IDs of the places that were recommended to the user. */
```
arielz-g: the list of places IDs that were recommended to the user

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+  public abstract ImmutableList<String> placesRecommendedToUser();
```
arielz-g: Can just be recommendedPlaces if the name is UserFeedback, think about the call UserFeedback.recommendedPlaces(), no need for repetition.

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /** @return the ID of the place the user chose, if any. */
+  public abstract Optional<String> placeUserChose();
```
arielz-g: Same here, may be chosenPlaces.

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Represents the feedback that users enter about the recommendations they were given.
+ */
+@AutoValue
+public abstract class UserFeedback {
+
+  /** @return a list of IDs of the places that were recommended to the user. */
+  public abstract ImmutableList<String> placesRecommendedToUser();
+
+  /** @return the ID of the place the user chose, if any. */
```
arielz-g, line 40: Why just the ID? Wouldn't it be better to have the entire Place object here and then extract the necessary information later on? Leaves you more room for changes IMO.

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+  public abstract Optional<String> placeUserChose();
+
+  /**
+  * @return true if the user requested to try again and get more recommendations (implying they
+  * weren't satisfied with the recommendations they recieved).
+  */
+  public abstract boolean userTriedAgain();
```
arielz-g, line 47: Can a user try again just once? What if the user clicked it multiple times (IMO that should be taken as a satisfactory factor looking forward), but even for now you can keep a "retriesCount" instead. WDYT?

project/src/main/java/com/google/sps/data/UserFeedback.java
arielz-g: Going over your discussion and looking at the implementation, I'm not sure it's indeed correct to validate this. If the UI enables this behavior, than it's not an invalid argument. Think about the user experience here, what would happen if the user picked a place now and then clicked try-again?

I believe this should first be handled in the UI (if this is what you want to enforce), i.e:
If the user picked a place and then clicked "try-again", if possible, treat it as if there are no choices made (this is something your implementation should handle, and shouldn't harm the user experience IMO).

On any case, whichever solution you opt for, I'd add the validation afterwards and not in this PR.

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Concludes the building of a new UserFeedback instance.
+     *
+     * @return the new instance.
+     * @throws IllegalArgumentException if the user feedback isn't consistent.
+     */
+    public UserFeedback build() throws IllegalArgumentException {
+      UserFeedback feedback = autoBuild();
+      ValidationUtils.validateChosenPlaceInReccomendedPlaces(feedback.placeUserChose(),
```
arielz-g: Use a static import.

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -16,38 +16,68 @@
...
+  /**
+   * @param placeUserChose the place the user chose to order from
+   * @param placesRecommendedToUser the places that the system recommended to the user
+   * @throws IllegalArgumentException if chosen place wasn't one of the recommended places
+   */
+  public static void validateChosenPlaceInReccomendedPlaces(Optional<String> placeUserChose,
+      ImmutableList<String> placesRecommendedToUser) throws IllegalArgumentException {
+    checkArgument(
+        !placeUserChose.isPresent() || placesRecommendedToUser.contains(placeUserChose.get()),
+        "Chosen place must be one of places that were recommended to the user."
```
arielz-g: "one of the" or "contained in the"

project/src/main/java/com/google/sps/data/ValidationUtils.java
```diff
@@ -16,38 +16,68 @@
...
+      boolean userTriedAgain) throws IllegalArgumentException {
+    checkArgument(
+        !(placeUserChose.isPresent() && userTriedAgain),
+        "User can't both try again and choose a place."
+    );
+  }
 
-    private ValidationUtils() {
-    }
+  private ValidationUtils() {
```
arielz-g: This can be a single line: private ValidationUtils() {}

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+@RunWith(JUnit4.class)
+public final class UserFeedbackTest {
+
+  private static final String PLACE_ID_1 = "12345";
+  private static final ImmutableList<String> PLACES_WITH_PLACE_ID_1 =
+      ImmutableList.of(PLACE_ID_1, "11111", "22222");
+
+  @Test
+  public void build_invalidBothChoosePlaceAndTryagain_throwsIllegalArgumentException() {
```
arielz-g: It's unnecessary to have a test for two invalid cases at once. As code execution goes, the first validation will throw and it won't reach the second one (ever).

The point in unit tests is to isolate events from one another, so if you have 3 validations, have 3 tests where the input is invalid at each test exactly where it "should".

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,82 @@
...
+      UserFeedback.builder()
+          .setPlacesRecommendedToUser(PLACES_WITH_PLACE_ID_1)
+          .setPlaceUserChose(PLACE_ID_1)
+          .setUserTriedAgain(true)
+          .build();
+    });
+  }
+
+  @Test
+  public void build_invalidChoosePlaceNotInRecommendedPlaces_throwsIllegalArgumentException() {
```
arielz-g: I'd do build_chosenPlaceNotInRecommendedPlaces_throwsIllegalArgumentException: the "invalid" here is redundant as name structure should be "method_scenario_outcome", and a "throwsIllegalArgumentException" makes it clear it's invalid:)

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -0,0 +1,82 @@
...
+      UserFeedback.builder()
+          .setPlacesRecommendedToUser(ImmutableList.of("11111", "22222", "33333"))
+          .setPlaceUserChose("44444")
+          .setUserTriedAgain(false)
+          .build();
+    });
+  }
+
+  @Test
+  public void build_validUserChosePlace_returnsValidUserFeedback() {
```
arielz-g: Too much "valid" in here (for the following test as well).
I'd do:
build_userChosePlace_validInstance
build_userTriedAgain_validInstance

karenav: done

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
karenav: done

project/src/main/java/com/google/sps/data/ValidationUtils.java
karenav: done

project/src/main/java/com/google/sps/data/ValidationUtils.java
karenav: done

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: done

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: I don't think this is information we would want to use for our current purposes, bc the number of times a user tried again says more about the user than about his thoughts about the specific current place. maybe we'll want to add this in the future but I don't think adding this now will help with the current purposes.

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: basically because we only need the ID for our current use of the data. I know that keeping more data would potentially allow more things, but what I understand from your past comments is that we shouldn't keep things that we don't intend to use (and also it might be wasteful). WDYT?

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: done

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: that's a great point, I didn't think about the context

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: done

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: done

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: I think it doesn't make sense to keep a listing of a place that is both chosen and tried again. I don't want to handle the FE in this PR, so I don't mind deleting it though I think it's a bit redundant to delete and potentially rewrite.
done.

project/src/main/java/com/google/sps/data/UserFeedback.java
arielz-g: That depends on the case: if you don't think you'll use the rest of the Place data looking forward, then it's ok, but if you spot a use case, I'd keep as is.

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -68,7 +67,7 @@ public boolean isRegistered(String userId) {
   * @param userId the id of the user that we want to register to our system
   */
   public void registerUser(String userId) {
-    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    ValidationUtils.validateNonEmptyString(userId, "Invalid user ID.");
```
arielz-g: This is a redundant change: the method has the exact same line, calling the method or having the checkArgument in place isn't adding anything new.

Note that the reason to have a Validations class is for specific validations that have logic in them (as there's already a Strings.isNullOrEmpty, you may simply use it:)).

project/src/main/java/com/google/sps/data/UserFeedback.java
```diff
@@ -78,11 +99,12 @@ public abstract Builder setPlacesRecommendedToUser(ImmutableList<String>
...
+      // The feedback time is set to be the time in which the object is created.
+      setFeedbackTimeInMillis(System.currentTimeMillis());
       UserFeedback feedback = autoBuild();
-      ValidationUtils.validateChosenPlaceInReccomendedPlaces(feedback.placeUserChose(),
-          feedback.placesRecommendedToUser());
-      ValidationUtils.validateUserTriedAgainOnlyIfdidntchoose(feedback.placeUserChose(),
-          feedback.userTriedAgain()); // TODO(M5): make sure UI supports this demand
+      validateChosenPlaceInReccomendedPlaces(feedback.chosenPlace(),
+          feedback.recommendedPlaces());
+      validateNonEmptyString(feedback.userId(), "User ID cannot be empty.");
```
arielz-g: Same here, keep it in place.

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
```diff
@@ -32,50 +32,39 @@
       ImmutableList.of(PLACE_ID_1, "11111", "22222");
 
   @Test
-  public void build_invalidBothChoosePlaceAndTryagain_throwsIllegalArgumentException() {
+  public void build_choosePlaceNotInRecommendedPlaces_throwsIllegalArgumentException() {
```
arielz-g: chosenPlace? Same for the second test.

project/src/main/java/com/google/sps/data/UserFeedback.java
karenav: done

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: done

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
arielz-g: Notice I wrote *chosen*, in here it says choose. I'm approving this PR but please fix before you push the changes.

arielz-g (approved):

project/src/test/java/com/google/sps/data/UserFeedbackTest.java
karenav: it was changed, it's not seen in the main screen because of Github - in the main screen you can't see the the new code, you just have an "outdated" mark at the top of the change request, that means the commented line was changed.

PR #59, opened 2020-12-21
Store user preferences

tbarnahor:
There is a small bug (in my opinion) that is not treated - every time a user presses try again if he doesn't like the results then his preferred cuisines are stored again. I think that they should be stored once on every session of "try again"s if they don't change. This can also be treated in the scoring (ignore near timestamps), just pointing out this issue.

### DIFF
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -70,7 +70,7 @@ public class DataAccessor {
   * @param userId the id of the user that we want to register to our system
   */
   public void registerUser(String userId) {
-    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    checkArgument(!Strings.isNullOrEmpty(userId), "User ID may not be null or empty");
     checkArgument(!isRegistered(userId), "User already registered.");
     Entity userEntity = new Entity(USER_ENTITY_KIND, userId);
     datastoreService.put(userEntity);
@@ -82,16 +82,14 @@ public class DataAccessor {
    * @param userId The ID of the user to store the preferred cuisines for.
    * @param userPref The user choices on the query form to store in the user’s database.
    */
-  public void storeUserPreferences(String userId, UserPreferences userPref) {
-    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
-    if (userPref.cuisines().isEmpty()) {
-      return;
+  public void storeUserPreferences(String userId, UserPreferences userPreferences) {
+    checkArgument(!Strings.isNullOrEmpty(userId), "User ID may not be null or empty");
+    if (!userPreferences.cuisines().isEmpty()) {
+      Entity prefsEntity = new Entity(PREFERNCES_ENTITY_KIND);
+      prefsEntity.setProperty("userId", userId);
+      prefsEntity.setProperty("date", new Date()); // TODO(Tal): Deal with try again sessions
+      prefsEntity.setProperty("preferedCuisines", userPreferences.cuisines());
+      datastoreService.put(prefsEntity);
     }
-    Entity prefsEntity = new Entity(PREFERNCES_ENTITY_KIND);
-    prefsEntity.setProperty("userId", userId);
-    prefsEntity.setProperty("date", new Date()); // TODO(Tal): Deal with try again sessions
-    prefsEntity.setProperty("preferedCuisines", userPref.cuisines());
-    datastoreService.put(prefsEntity);
   }
-
 }
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -79,7 +79,6 @@ public class DataAccessor {
 
   /**
    * Stores the UserPreferences in the personalized user database.
-   * If no cuisines were prefered by the user, nothing is stored.
    *
    * @param userId The ID of the user to store the preferred cuisines for.
    * @param userPref The user choices on the query form to store in the user’s database.
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -79,12 +79,16 @@ public class DataAccessor {
 
   /**
    * Stores the UserPreferences in the personalized user database.
+   * If no cuisines were prefered by the user, nothing is stored.
    *
    * @param userId The ID of the user to store the preferred cuisines for.
    * @param userPref The user choices on the query form to store in the user’s database.
    */
   public void storeUserPreferences(String userId, UserPreferences userPref) {
     checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    if (userPref.cuisines().isEmpty()) {
+      return;
+    }
     Entity userEntity = new Entity(PREFERNCES_ENTITY_KIND);
     userEntity.setProperty("userId", userId);
     userEntity.setProperty("date", new Date());
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -15,6 +15,7 @@
 package com.google.sps.data;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
 
 import java.util.Date;
 import com.google.common.annotations.VisibleForTesting;
@@ -27,7 +28,6 @@ import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.repackaged.com.google.api.client.util.Strings;
 
 public class DataAccessor {
 
@@ -54,7 +54,7 @@ public class DataAccessor {
   *     previously added to datastore.
   */
   public boolean isRegistered(String userId) {
-    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    checkArgument(!Strings.isNullOrEmpty(userId), "User ID may not be null or empty");
     Key userIdKey = KeyFactory.createKey(USER_ENTITY_KIND, userId);
     Filter userIdFilter =
         new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, userIdKey);
--- a/project/src/main/java/com/google/sps/data/DataAccessor.java
+++ b/project/src/main/java/com/google/sps/data/DataAccessor.java
@@ -16,6 +16,8 @@ package com.google.sps.data;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import java.util.Date;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
@@ -32,7 +34,8 @@ public class DataAccessor {
 
   private final DatastoreService datastoreService;
   @VisibleForTesting
-  static final String USER_ENTITY_NAME = "User";
+  static final String USER_ENTITY_KIND = "User";
+  static final String PREFERNCE_ENTITY_KIND = "UserPreferences";
 
   /**
    * A constructor that creates a DatastoreService instance for the class.
@@ -53,10 +56,10 @@ public class DataAccessor {
   */
   public boolean isRegistered(String userId) {
     checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
-    Key userIdKey = KeyFactory.createKey(USER_ENTITY_NAME, userId);
+    Key userIdKey = KeyFactory.createKey(USER_ENTITY_KIND, userId);
     Filter userIdFilter =
         new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, userIdKey);
-    Query query = new Query(USER_ENTITY_NAME).setFilter(userIdFilter).setKeysOnly();
+    Query query = new Query(USER_ENTITY_KIND).setFilter(userIdFilter).setKeysOnly();
     return datastoreService.prepare(query)
         .asList(FetchOptions.Builder.withDefaults())
         .size() > 0;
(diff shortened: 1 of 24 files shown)


project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -16,6 +16,8 @@
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import java.util.Date;
+
```
karenav: extra space

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -70,7 +73,26 @@ public boolean isRegistered(String userId) {
...
+   *
+   * @param userId The ID of the user to store the preferred cuisines for.
+   * @param userPref The user choices on the query form to store in the user’s database.
+   */
+  public void storeUserPreferences(String userId, UserPreferences userPref) {
+    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    if (userPref.cuisines().isEmpty()) {
+      return;
+    }
+    Entity userEntity = new Entity(PREFERNCES_ENTITY_KIND);
```
karenav: I think the naming isn't accurate, maybe prefEntity is more suitable

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -98,4 +115,15 @@ private static LatLng getLatLngFromString(String coordinates) {
...
   }
+
+  // Store the user's preferences in the database, only if the user is signed in an registered.
+  private void storePreferences(String userIdToken, UserPreferences userPrefs) {
+    Optional<String> optionalUserId = userVerifier.getUserIdByToken(userIdToken);
+    if (!optionalUserId.isPresent()) {
+        return;
+    }
+    String userId = optionalUserId.get();
+    System.out.println(userId);
```
karenav: notice the print

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -98,4 +115,15 @@ private static LatLng getLatLngFromString(String coordinates) {
     String[] latLng = coordinates.split(",");
     return new LatLng(Double.parseDouble(latLng[0]), Double.parseDouble(latLng[1]));
   }
+
+  // Store the user's preferences in the database, only if the user is signed in an registered.
```
karenav: I think there's a typo - "and" instead of "an".
I think it's not accurate to say that this happens if the user is signed in *and registered*, because you don't check if they're registered (you assume that, which I think is fine).

project/src/main/webapp/index.html
```diff
@@ -23,18 +23,18 @@ <h1> FoodMood - your food order helper!</h1>
     <div id="query-form" class="query-form">
       <form id="cuisines-form" class="form-element-toolbar">
         <p>Select cuisines:</p>
-        <input type="checkbox" id="cuisine-italian" name="cuisine" value="italian" checked>
```
karenav, line 26: why would we not want all these to be checked? I think it's easier for the user to keep all of them checked as default

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -33,6 +33,9 @@
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
 import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.common.collect.ImmutableList;
+import com.google.maps.model.LatLng;
+
```
karenav: I think extra space

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -29,6 +30,8 @@
 import javax.servlet.http.HttpServletResponse;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.util.Optional;
+
```
karenav: I think extra space

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -176,11 +183,41 @@ public void getRequest_placesAndUserLocationForwadedToScorer() throws Exception
...
-    servlet.doGet(REQUEST, RESPONSE);
+    servlet.doPost(REQUEST, RESPONSE);
 
     verify(SCORER).getScores(places, new LatLng(00, 00));
   }
 
+  @Test
+  // This test checks that storeUserPreferences is called with the expected parameters
+  // when the servlet gets a valid ID token
+  public void getRequest_validIdToken_userIdAndPreferencesForwardedForStoring() throws Exception {
```
karenav: If it's now a post request, the function naming is inaccurate

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -176,11 +183,41 @@ public void getRequest_placesAndUserLocationForwadedToScorer() throws Exception
...
+    initializeRequestParameters();
+    when(FETCHER.fetch(any(UserPreferences.class))).thenReturn(ImmutableList.of());
+    when(REQUEST.getParameter("idToken")).thenReturn("token");
+    when(USER_VERIFIER.getUserIdByToken("token")).thenReturn(Optional.of("userId"));
+
+    servlet.doPost(REQUEST, RESPONSE);
+
+    UserPreferences expectedUserPrefs = getValidUserPreferencesBuilder().build();
+    verify(DATA_ACCESSOR).storeUserPreferences("userId", expectedUserPrefs);
```
karenav, line 201: The only thing that bothers me is that the two places that update the preferences are assumed to do the exact same thing, but they are independent and I think that's a problem. I'll explain: the function "initializeRequestParameters" initializes the preferences to be exactly the same as the preferences in "getValidUserPreferencesBuilder()", so this test works. but these two are independent, which means someone can come and change one of them a bit, for example change the rating to 5, and that wouldn't be a wrong thing to do, but it will cause this test to fail. 
I suggest an "ugly" solution, which in my opinion is more safe, which is to set the attributes inside the test, so that it's clear that changing something may effect the test results. Either do what the functions do (which is uglier), or make the functions receive input and send the same input to both, or something like that.

karenav (changes requested):
Looks really great! 
About the "bug" you mentioned - I'm not sure It's a bug, because if the user changes their preferences after trying again, we do want to keep track of the new prefs they entered as well. 
As you said. maybe the right thing to do is give different identical prefs of the same day less weight (so if the user had the same prefs in two different days, it has more affect than if they had the same prefs twice on the same day, which has a bit more affect than if they had the prefs once).

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: Totaly!

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: Right

project/src/test/java/com/google/sps/data/DataAccessorTest.java
tbarnahor: Right

project/src/main/webapp/index.html
tbarnahor: I think it is not comftable for the user (I actually changed it after running our app a few times and found it very annoying to delete all checks except one or two every time).
I changed the title so maybe it is more clear to the user now that if he doesn't want to limit the search then he doesn't need to pick anything. 
Anyway this doesn't have to do with this PR so it's my bad that this is here. Lets discuss UX next week for final touchups.

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: Sure, I changed it

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: good catch

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: Right, it remained from your method I copied haha

project/src/main/webapp/index.html
karenav: we should discuss that. does that match the program? if they choose nothing they get everything? from my memory if they choose nothing they get an error

tbarnahor: Yes I removed the error that forces to check a cuisine in another PR that I was supposed to push and it got postponed. I added it now so we have it all here. Other then that the fetcher has no problem getting 0 prefered cuisines, it returns all sort of cuisines in this case (as we discussed) so it works fine, and IMO better.

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: Ok, I changed it as you suggested

karenav (approved):

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -70,7 +72,26 @@ public boolean isRegistered(String userId) {
...
     datastoreService.put(userEntity);
   }
+
+  /**
+   * Stores the UserPreferences in the personalized user database.
+   *
+   * @param userId The ID of the user to store the preferred cuisines for.
+   * @param userPref The user choices on the query form to store in the user’s database.
+   */
+  public void storeUserPreferences(String userId, UserPreferences userPref) {
```
arielz-g: nit: use full name, for the argument (userPreferences).

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -70,7 +72,26 @@ public boolean isRegistered(String userId) {
...
+    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
+    if (userPref.cuisines().isEmpty()) {
```
arielz-g: I'd do:
if(!userPref.cuisines().isEmpty()){// logic)}

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -70,7 +72,26 @@ public boolean isRegistered(String userId) {
...
   }
+
+  /**
+   * Stores the UserPreferences in the personalized user database.
+   *
+   * @param userId The ID of the user to store the preferred cuisines for.
+   * @param userPref The user choices on the query form to store in the user’s database.
+   */
+  public void storeUserPreferences(String userId, UserPreferences userPref) {
+    checkArgument(!Strings.isNullOrEmpty(userId), "Invalid user ID");
```
arielz-g: nit: "User ID may not be null or empty"

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -70,7 +72,26 @@ public boolean isRegistered(String userId) {
...
+    if (userPref.cuisines().isEmpty()) {
+      return;
+    }
+    Entity prefsEntity = new Entity(PREFERNCES_ENTITY_KIND);
```
arielz-g: Can't this whole thing be in-line?

datastoreService.put(new Entity(PREFERNCES_ENTITY_KIND).setProperty(..).setProperty(..));

project/src/main/java/com/google/sps/servlets/QueryServlet.java
```diff
@@ -70,6 +83,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
               .setLocation(getLatLngFromString(request.getParameter("location")))
               .setCuisines(ImmutableList.copyOf(request.getParameter("cuisines").split(",")))
               .build();
+      String userIdToken = request.getParameter("idToken");
+      if (!userIdToken.isEmpty()) {
```
arielz-g: You can move this condition to storePreferences.

project/src/test/java/com/google/sps/data/DataAccessorTest.java
```diff
@@ -125,13 +129,66 @@ public void registerUser_alreadyRegistered_throwIllegalArgumentException() {
...
+      userId, getValidUserPreferencesBuilder().setCuisines(CUISINES).build());
+    List<Entity> results = createPreparedQueryByUserIdProperty(userId)
+        .asList(FetchOptions.Builder.withDefaults());
+
+    assertEquals(1, results.size());
+    assertEquals(CUISINES, results.get(0).getProperty("preferedCuisines"));
+  }
+
+  @Test
+  public void storeUserPreferences_validUserIdNoPreferedCuisines_nothingStored() {
```
arielz-g: nit: Preferred

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
```diff
@@ -164,23 +173,62 @@ public void getRequest_userPreferencesForwardedToFetcher() throws Exception {
...
+    servlet.doPost(REQUEST, RESPONSE);
 
     verify(SCORER).getScores(places, new LatLng(00, 00));
   }
 
+  @Test
+  // This test checks that storeUserPreferences is called with the expected parameters
+  // when the servlet gets a valid ID token
+  public void postRequest_validIdToken_userIdAndPreferencesForwardedForStoring() throws Exception {
+    when(REQUEST.getParameter("rating")).thenReturn("4");
```
arielz-g: You already have initializeRequestParameters called in the before method, why reset this here?

If it's for clarity as for when you're building the UserPreferences object (which I get), you can have all of these variants as constants & have a method that returns the matching UserPreferences object.

arielz-g (approved):
Great PR!

Approving this as these are all minor comments overall. You're welcome to ask me if something's unclear:)

project/src/test/java/com/google/sps/servlets/QueryServletTest.java
tbarnahor: Yes it was for clarity :)
I changed it as you suggested

project/src/main/java/com/google/sps/servlets/QueryServlet.java
tbarnahor: Good idea

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: This is not possible because setProperty returns void, so it all has to be set seperately.

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: Sure

project/src/main/java/com/google/sps/data/DataAccessor.java
tbarnahor: I like it better too :)

PR #60, opened 2020-12-28
Places Fetcher requests generators

tbarnahor:
This PR adds a places details request generator and a text search generator, both with production and fake implementations, that are passed to the fetcher. The purpose of this change is to be able to check the arguments of a request in the tests and act accordingly. This testing ability is especially needed for the next PR, which deals with sending a seperate request for each cuisine in order to store the place's cuisine.


project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
```diff
@@ -0,0 +1,21 @@
+package com.google.sps.data;
```
szwbar: Why is this not in the test directory? (same for similar classes)

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
```diff
@@ -0,0 +1,21 @@
...
+    public String placeId;
+
+    /**
+     * FakeTestSearchRequest constructor.
+     *
+     * @param geoApiContext the GeoApiContext used for all Google GEO API requests
+     * @param placeId the Place ID to retrieve details for
+     */
+    public FakePlaceDetailsRequest(GeoApiContext context, String placeId) {
+        super(context);
```
szwbar: Why do you need the context here?

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
szwbar: This can also reside as a nested class of the fake generator, instead of a class of its own.

project/src/main/java/com/google/sps/data/FakeSearchRequest.java
```diff
@@ -0,0 +1,22 @@
...
+    public String searchWords;
+
+    /**
+     * FakeTestSearchRequest constructor.
+     *
+     * @param geoApiContext the GeoApiContext used for all Google GEO API requests
+     * @param searchWords the search words used for the text search
+     */
+    public FakeSearchRequest(GeoApiContext context, String searchWords) {
+        super(context);
```
szwbar: Same question..

project/src/main/java/com/google/sps/data/FakeSearchRequest.java
```diff
@@ -0,0 +1,22 @@
+package com.google.sps.data;
+
+import com.google.maps.TextSearchRequest;
+import com.google.maps.GeoApiContext;
+
+public class FakeSearchRequest extends TextSearchRequest {
+
+    /** The search wirds used for the text search */
```
szwbar: typo: words

project/src/main/java/com/google/sps/data/FakeSearchRequestGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+    // The entry point for a Google GEO API request.
+    private GeoApiContext context;
+
+    /**
+     * FakeSearchRequestGenerator constructor.
+     *
+     * @param geoApiContext the GeoApiContext used for all Google GEO API requests
+     */
+    public FakeSearchRequestGenerator(GeoApiContext geoApiContext) {
+        this.context = geoApiContext;
```
szwbar: Same question..

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -55,8 +53,11 @@
     // The maximal number of times the search radius will be extended.
     private static final int MAX_NUM_OF_RADIUS_EXTENSIONS = 4;
 
-    // The entry point for a Google GEO API request.
-    private GeoApiContext context;
+    // The generator of TextSearchRequests.
```
szwbar: nit: A generator for...

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -68,10 +69,16 @@
...
      *
-     * @param geoApiContext the GeoApiContext used for all Google GEO API requests
+     * @param textSearchRequestGenerator
+     *     used for generating the TextSearchRequests sent to Google Places API
+     * @param placeDetailsRequestGenerator
+     *     used for generating the PlaceDetailsRequests sent to Google Places API
      */
-    public PlacesFetcher(GeoApiContext geoApiContext) {
-        this.context = geoApiContext;
+    public PlacesFetcher(
```
szwbar: I think that you should accept those in the @VisibleForTesting c'tor only. 
Then you have one c'tor same as you had before, and one test c'tor that provides you the fake generators. 
The non-test c'tor creates the generators itself, simplifying the calling code (QueryServlet does not need to know how SearchRequestGenerator and PlaceDetailsGenerator are). 
WDYT?

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
```diff
@@ -0,0 +1,21 @@
...
+import com.google.maps.PlaceDetailsRequest;
+import com.google.maps.GeoApiContext;
+
+public class FakePlaceDetailsRequest extends PlaceDetailsRequest {
+
+    /** The Place ID to retrieve details for. */
+    public String placeId;
+
+    /**
+     * FakeTestSearchRequest constructor.
```
szwbar: typo: FakePlaceDetailsRequest

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -154,8 +158,23 @@ private static PlacesSearchResult createTestPlacesSearchResult(String placeId) {
     return searchResult;
   }
 
+  private static ArgumentMatcher<FakeSearchRequest> matchExpectedSearchRequest(
```
szwbar: nit: matchesSearchRequest

* Expected is redundant here
* matches is usually combined in a sentence-like statement.

szwbar (changes requested):
The most important thing is to move all the fakes into the test that uses them. They shouldn't be in the production code. 
Other than that, looks great!

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
tbarnahor: I didn't think about, it does make much more sense to put it in the test dierctory :)
I added the fake requests as public static nested classes - static because they are not related to an instance of the generator, and public because they are accessed from the tester. Let me know if this is fine.

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
tbarnahor: Because the parent's constructor takes this as a parameter (it has no default parameter-less constructor), so I need to call super() with this argument. I did wonder if I should create some dummy context to pass here instead. What do you think?

project/src/main/java/com/google/sps/data/FakeSearchRequest.java
tbarnahor: Same answer..

project/src/main/java/com/google/sps/data/FakeSearchRequestGenerator.java
tbarnahor: This is so the contex is ready to be passed to the FakeSearchRequest constructor on every create() call.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Very good idea. It bothered me too that the queryServlet is exposed to them.

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
tbarnahor: Oh no you are right

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: makes sense

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
szwbar: 👍

project/src/main/java/com/google/sps/data/FakePlaceDetailsRequest.java
szwbar: Oh I forgot that in Java you have to call super... OK so it's fine as-is.

szwbar (approved):
LGTM1

karenav (approved):

PR #61, opened 2020-12-29
Take user feedback

karenav:


project/src/main/webapp/index.html
```diff
@@ -13,81 +13,81 @@
...
-        <label for="device-location">Press here to use your device's location!</label>
-        <div id="map-warning-container">Please note: Mac devices might display an inaccurate location.
-          <br> It is recommended for Mac users to use the search box.</div>
-      </form>
-      <input id="location-input" class="controls" type="text" placeholder="Search Box"/>
-      <div id="map-error-container" class="error-text"></div>
-      <div id="map"></div>
-      <script> addMapWithSearchBox() </script>
-     </div>
+      <div id="sign-in-segment">
```
tbarnahor: I don't know why this happened, but almost all lines are marked as changed while I think only a few lines were added. It is pretty hard to review this way so maybe if you can revert to the last version and just add the lines and tabs that you wanted to add in this PR it can be great.

If it's too much time to do that now, you can leave it like this, but can you point out the changes? Only "user-input" class was added right?

project/src/main/webapp/index.html
```diff
@@ -100,18 +100,22 @@ <h2>Food Mood's recommendation:</h2>
...
         <label for="liking-feedback">Like these options?<br> Let us know what you chose: <br></label>
-        <select class="button" name="feedback" id="feedback">
-          <option value="1">First option</option>
-          <option value="2">Second option</option>
-          <option value="3">Third option</option>
-          <br>
-          <input type="submit" value="Submit" style="font-weight: bold">
-        </select>
+        <form id="chosen-place-form" class="form-element-toolbar">
+          <input type="radio" id="option-1" value="1" checked="checked">
```
tbarnahor: Why is this checked already? It can make this be set as chosen while the user didn't actively choose this.
Also, just wondering - why did you change to radio rather then the drop down we had?

project/src/main/webapp/script.js
```diff
@@ -395,3 +402,29 @@ function registerUserByToken() {
...
+      'Please select the place that you chose.');
+  if (indexOfPlaceUserChose <= recommendedPlaces.length) {
+    updateUserFeedback(recommendedPlaces, recommendedPlaces[indexOfPlaceUserChose].placeId,
+        false /** user tried again */)
+  }
+}
+
+/** Updates user feedback in the database, by using the feedback servlet. */
+function updateUserFeedback(recommendedPlaces, chosenPlaceId, tryAgain) {
+  if (!googleUser) {
```
tbarnahor: I got this tip from Ariel on my PR and agreed with it:
write:

if (googleUser){
  logic
}

And then no "return" is needed.

tbarnahor (commented):
Nice job

project/src/main/webapp/script.js
karenav: That's weird, I think a while ago I got the opposite tip from her, and the explanation was that we don't want to put the big chunk of code inside an "if". why do you think this is better?

project/src/main/webapp/index.html
karenav: Radio - only because that's the way I know how to read the chosen property in the javascript. if you prefer it as a scroll list I think it's possible, just would take a little time to make sure we take the information correctly, I preffered to spare the time.
It's checked for convenience.. not a critical point but I think it's nicer that something is already checked. either way sending the feedback is an active action, because the user has to press "submit"

project/src/main/webapp/index.html
karenav: It happened because the big divs changed. sadly even if we revert it, it will still look the same here, because git doesn't handle html very well.

tbarnahor: Oh ok probably because it's html

project/src/main/webapp/index.html
tbarnahor: Ok I agree we don't have the time for these front end stuff

project/src/main/webapp/script.js
tbarnahor: I think it's more elegant without the "return". But there is no right answer..

szwbar: My 2 cents: I actually think differently than the first comment suggested. 
This code is nested:
```
if (condition) {
   // Do something very long which might have other conditions 
  // and for loops etc.
}
```

But this code is less nested:
```
if (!condition) {
  // nothing to do
  return;
}
// The rest of the code is at the upper level, causing 
// less confusion for the reader who needs to know at 
// what level of nesting the code currently is
```

This is actually one of the C++ best practices at Google.

project/src/main/webapp/script.js
```diff
@@ -395,3 +402,29 @@ function registerUserByToken() {
...
+  let indexOfPlaceUserChose = getCheckedValueByElementId('chosen-place-form',
+      'Please select the place that you chose.');
+  if (indexOfPlaceUserChose <= recommendedPlaces.length) {
+    updateUserFeedback(recommendedPlaces, recommendedPlaces[indexOfPlaceUserChose].placeId,
+        false /** user tried again */)
+  }
+}
+
+/** Updates user feedback in the database, by using the feedback servlet. */
+function updateUserFeedback(recommendedPlaces, chosenPlaceId, tryAgain) {
```
szwbar: Why do you pass `recommendedPlaces` as a function param if it's a global variable?

tbarnahor (approved):

project/src/main/webapp/script.js
```diff
@@ -395,3 +402,29 @@ function registerUserByToken() {
...
   fetch('/register?idToken=' + googleUser.getAuthResponse().id_token, {method: 'POST'});
 }
+
+/** Sends feedback that includes the place the user chose. */
+function sendUserChoiceAsFeedback() {
+  let indexOfPlaceUserChose = getCheckedValueByElementId('chosen-place-form',
+      'Please select the place that you chose.');
+  if (indexOfPlaceUserChose <= recommendedPlaces.length) {
+    updateUserFeedback(recommendedPlaces, recommendedPlaces[indexOfPlaceUserChose].placeId,
+        false /** user tried again */)
```
szwbar, line 444: Use the variable name in the comment:
```
false /* tryAgain */)
```

szwbar (approved):
LGTM

project/src/main/webapp/script.js
karenav: Good point, fixed

project/src/main/webapp/script.js
karenav: Good point, done

PR #64, opened 2021-01-04
Add user attribute to feedback (when using) - tiny PR

karenav:


tbarnahor (approved):

project/src/main/java/com/google/sps/data/DataAccessor.java
```diff
@@ -54,7 +54,7 @@ public DataAccessor() {
   *     previously added to datastore.
   */
   public boolean isRegistered(String userId) {
-    checkArgument(!Strings.isNullOrEmpty(userId), "User ID may not be null or empty");
+    checkArgument(!isNullOrEmpty(userId), "User ID may not be null or empty");
```
szwbar, line 57: Is this change intended?

szwbar (approved):
Note that your PR description is incorrect or at least lacking, since you added this only in tests.

project/src/main/java/com/google/sps/data/DataAccessor.java
karenav: Yes

PR #65, opened 2021-01-05
Add cuisines field to place

tbarnahor:


project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
+            for (String cuisine: cuisines) {
+                attemptsCounter++;
+                try {
+                    resultsForCuisine = getPlacesSearchResults(genTextSearchRequest(
+                        preferences, INIT_SEARCH_RADIUS_M * attemptsCounter, cuisine));
+                } catch (ApiException
+                        | InterruptedException
+                        | IOException
+                        | IllegalStateException e) {
+                    throw new FetcherException("Couldn't fetch places from Places API", e);
```
karenav, line 126: Something to think about - if one cuisine fails that means we want to throw an error for the entire request? 
I think maybe if we get some results without one of the cuisines it's better than eliminating the whole request (We can also inform the user about which cuisines if we want, of course not necessarily in this PR)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
+                }
+                for (PlacesSearchResult result : resultsForCuisine) {
+                    placesSearchResults.computeIfAbsent(
+                        result.placeId, k -> new HashSet<>()).add(cuisine);
+                }
```
karenav, line 132: It looks really good!
I have a point I'm not sure about - from what I understand about computeIfAbsent, if the key already exists, it does nothing. So doesn't that mean a single place can only be added to the HashMap with one cuisine?

project/src/test/java/com/google/sps/data/PlaceTest.java
```diff
@@ -17,7 +17,10 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+
```
karenav: extra space

project/src/test/java/com/google/sps/data/PlaceTest.java
```diff
@@ -17,7 +17,10 @@
...
 import com.google.maps.model.LatLng;
+import com.google.common.collect.ImmutableList;
+
```
karenav: I think the static imports should be before the non static ones

karenav (approved):

project/src/main/java/com/google/sps/data/Place.java
```diff
@@ -143,6 +149,12 @@ public static Builder builder() {
      */
     public abstract Builder setBusinessStatus(BusinessStatus businessStatus);
 
+    /**
+     * @param cuisines the place's cuisines.
+     * @return a Place builder that enables to continue building
+     */
+    public abstract Builder setCuisines(ImmutableList<String> cuisines);
```
arielz-g: Nice to know (not a must), if you set the method's signature to:
public abstract Builder setCuisines(String... cuisines);

AutoValue well automatically set up a method that knows to deal with a list input as well as varargs (makes tests nicer when you set few values).

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
      *     for places or for places details
      */
     public ImmutableList<Place> fetch(UserPreferences preferences) throws FetcherException {
-        PlacesSearchResult[] placesSearchResult;
+        Map<String, Set<String>> placesSearchResults = new HashMap<>();
```
arielz-g: Use an ImmutableMap & ImmutableSet (If you wish to insert use a Builder in instantiation).

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
+        PlacesSearchResult[] resultsForCuisine;
+        // If user didn't choose any cuisines, search on all possible cuisines
+        Collection<String> cuisines =
```
arielz-g: Use the ImmutableMap, no need for Collection.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
         } while (
-            placesSearchResult.length < MIN_NUM_OF_RESULTS
+            placesSearchResults.size() < MIN_NUM_OF_RESULTS
             && attemptsCounter < MAX_NUM_OF_RADIUS_EXTENSIONS);
-        return createPlacesList(placesSearchResult);
+        return createPlacesList(placesSearchResults);
     }
 
-    private TextSearchRequest genTextSearchRequest(UserPreferences preferences, int radius) {
+    private TextSearchRequest genTextSearchRequest(
```
arielz-g: generateTextSearchRequest / createTextSearchRequest.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
-            attemptsCounter++;
-            try {
-                placesSearchResult = getPlacesSearchResults(
-                    genTextSearchRequest(preferences, INIT_SEARCH_RADIUS_M * attemptsCounter));
-            } catch (ApiException | InterruptedException | IOException | IllegalStateException e) {
-                throw new FetcherException("Couldn't fetch places from Places API", e);
+            for (String cuisine: cuisines) {
+                attemptsCounter++;
+                try {
+                    resultsForCuisine = getPlacesSearchResults(genTextSearchRequest(
```
arielz-g: nit: cuisineResults.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
```diff
@@ -104,25 +107,40 @@ public PlacesFetcher(GeoApiContext geoApiContext) {
...
+                ? CUISINE_TO_SEARCH_WORDS.keySet() : preferences.cuisines();
         int attemptsCounter = 0;
         do {
-            attemptsCounter++;
-            try {
-                placesSearchResult = getPlacesSearchResults(
-                    genTextSearchRequest(preferences, INIT_SEARCH_RADIUS_M * attemptsCounter));
-            } catch (ApiException | InterruptedException | IOException | IllegalStateException e) {
-                throw new FetcherException("Couldn't fetch places from Places API", e);
+            for (String cuisine: cuisines) {
```
arielz-g, line 117: This whole bit is a tad long and could be made more readable IMO. Consider the following:

1. Export the entire try / catch bit to getPlacesSearchResults and make it throw the FetcherException.
2. Use a stream, something like:

placesSearchResults = cuisines.stream().map(cuisine -> getPlacesSearchResults(..)).collect(toImmutableHashMap(cuisineResult::placeId, cuisineResult -> toImmutableHashSet(..));

Note that this is a draft and I probably missed a bit, but I recommend you check out Stream's flatMap and the collectors for all collection types (toImmutable<type>). Also, there's a groupingBy collector I think will be useful here: https://mkyong.com/java8/java-8-collectors-groupingby-and-mapping-example/. The bottom line is this whole thing (up to the while part) may be a one liner IMO.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
```diff
@@ -72,7 +71,7 @@
   private static final String PLACEID_2 = "ChIJ02qnq0KuEmsRHUJF4zo1x4I";
 
   /** Valid Place objects. */
-  private static final Place PLACE_1 =
+  private static final Place.Builder PLACE_1_BUILDER =
```
arielz-g: If you wish to use a Builder object, use a method and not a static final (createValidPlaceBuilder()), or better yet: createValidPlace(String... cuisines) - no builder, use the varargs whenever you need to.

project/src/test/java/com/google/sps/data/PlaceTest.java
tbarnahor: Right

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: No, that's exactly why I chose to use this - so a single place can have multiple cuisines :)
It's working and tested.
The computeIfAbsent creates a new Set of cuisines if the result wasn't already found, but if the results already exists then its cuisines set is returned and the new cuisine is added to it.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: That's a really good point, actually it's relevant before this PR also because it could've failed in one iteratoin in the radius extensoions and nothing would be returned. I will add this as a TODO and try treating this in another PR.

project/src/test/java/com/google/sps/data/PlacesFetcherTest.java
tbarnahor: Much better :)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I agree it's better

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Changed it :)

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: Right!

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: I changed to immutableSet builder
But I am having trouble doing that with the map because I need to have the ability to get objects from it in order to add a cuisine to a place already in the map. 
So is it fine to work with a regular map and then when passing it the the next method pass an immutable copy?

project/src/main/java/com/google/sps/data/Place.java
tbarnahor: I started to change it but then I realized that the tests are already built with a cuisines immutable list and that it is needed this way for other uses, so this change will require to change more classes. I will keep it mind for next times because I see how it simplifies the tests.

project/src/main/java/com/google/sps/data/PlacesFetcher.java
tbarnahor: 1. I didn't want to have any logic in the getPlacesSearchResults because this method is not tested (it's always stubed, as a solution for the query.await() method from the Places API library being untestable). So I think it is better to leave it that way regarding the try/ catch.

2. I tried different options using streams but it all seemed less readable to me.
The main problems were that:
* I need to increment the attemptsCounter for every cuisine in the stream, and the only way I found to do that is by using AtomicInteger (which I wanted to know what you think about that in general)
* I need to be able to know for each result what was the cuisine that mapped that result. So the map method will need to generate something more sophisticated then just the result, because once I use it I loose that information.
Both of these points made the .map() in the stream to have a lot of logic that I think makes it less readable then now.

We can discuss this 1:1 if you have the time :)

arielz-g (approved):

PR #66, opened 2021-01-05
Feedback servlet

karenav:


project/src/main/webapp/script.js
```diff
@@ -400,3 +407,29 @@ function registerUserByToken() {
...
+  let indexOfPlaceUserChose = getCheckedValueByElementId('chosen-place-form',
+      'Please select the place that you chose.');
+  if (indexOfPlaceUserChose <= recommendedPlaces.length) {
+    updateUserFeedback(recommendedPlaces, recommendedPlaces[indexOfPlaceUserChose].placeId,
+        false /** user tried again */)
+  }
+}
+
+/** Updates user feedback in the database, by using the feedback servlet. */
+function updateUserFeedback(recommendedPlaces, chosenPlaceId, tryAgain) {
```
tbarnahor: We already have a "getUserIdToken" function in our js, you should use it instead of lines 423-425. And do :
      `idToken=${getUserIdToken()}`
Just like the fetchFromQuery request.

Notice that this way the check in the Servlet should be:
If `userIdToken.isEmpty()` instead of `userIdToken == null`, just like the treatment in the Query Servlet.
I think it is also better for consistency to have it the same.

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,83 @@
...
+  @VisibleForTesting
+  void init(UserVerifier userVerifier, DataAccessor dataAccessor) {
+    this.userVerifier = userVerifier;
+    this.dataAccessor = dataAccessor;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userIdToken = request.getParameter("idToken");
+    if (userIdToken == null) {
```
tbarnahor: See my comment on the js - for consistency I think idToken paramater should be passed in the same manner it is passed to the Query Servlet, and then this line should be:  `if (userIdToken.isEmpty())`

project/src/main/webapp/script.js
tbarnahor: Also I think the name of this function is a little confusing. It stores the feedback right? It doesn't update a feedback already existing. So maybe change to storeUserFeedback.
Just a suggestion, do whatever you think.

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Test
+  public void doPost_validToken_updatesFeedback() throws Exception {
+    mockValidRequestParams();
+    when(mockUserVerifier.getUserIdByToken(ID_TOKEN)).thenReturn(Optional.of(USER_ID));
+
+    servlet.doPost(REQUEST, RESPONSE);
+
+    // We are unable to make sure the exact function input, because we can't create an identical
```
tbarnahor: You can check that the other fields are correct with argument matcher (argThat).
If you are interested in doing that check you can look in the Places Fetcher test for an example of how I used it to check that the expected instance is passed.

tbarnahor (approved):
Looks good!
I am approving so I don't hold you back.
Let me know if you have questions about the testing suggestion.

project/src/main/webapp/script.js
karenav: That's right (we have the function), but we don't want to get rid of the check (423-425), because we don't want to call the servlet if we don't have a user (there is no reason to add a redundant call, even if the case is checked)
And I now made both servlets check for both things :)

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: we now check for both things, done

project/src/main/webapp/script.js
karenav: Great suggestion, done

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
karenav: That's a great technique that I wasn't aware of! Thanks for letting me know, definitely used it

project/src/main/webapp/script.js
tbarnahor: That check was a part of the function... the exact same lines 
There is no need to change anything in the Query Servlet

karenav: (You want the current function to return, it's not enough that the function you call returns, if you just call the function and not do the check you'll still have a redundant call to the servlet)

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.data.DataAccessor;
+import com.google.sps.data.UserFeedback;
+import com.google.sps.data.UserVerifier;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+
+/** A servlet that registers a user, according to a given token. */
```
szwbar: Update comment.

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,84 @@
...
+
+
+  @Override
+  public void init() {
+    this.userVerifier = UserVerifier.create(System.getenv("CLIENT_ID"));
+    this.dataAccessor = new DataAccessor();
+  }
+
+  @VisibleForTesting
+  void init(UserVerifier inUserVerifier, DataAccessor inDataAccessor) {
```
szwbar: why the `in` prefix?

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,84 @@
...
+    String userIdToken = request.getParameter("idToken");
+    if (isNullOrEmpty(userIdToken)) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, INVALID_TOKEN_MSG);
+      return;
+    }
+    Optional<String> optionalUserId = userVerifier.getUserIdByToken(userIdToken);
+    if (!optionalUserId.isPresent()) {
+      response.sendError(HttpServletResponse.SC_NOT_FOUND, NO_USER_TOKEN_MSG);
+      return;
+    }
```
szwbar: Why did you decide not to move this into some helper class? 
All of this logic is shared with the registration servlet. 
You can have something like:
```
if (!TokenValidator.validateRequest(request, response)) {
  // Token returns false and modifies the response if an invalid argument was given. 
  return;
}
```

project/src/main/webapp/index.html
```diff
@@ -111,18 +112,22 @@ <h2>Food Mood's recommendation:</h2>
...
+      <div id="feedback-box" style="display: none">
         <label for="liking-feedback">Like these options?<br> Let us know what you chose: <br></label>
-        <select class="button" name="feedback" id="feedback">
-          <option value="1">First option</option>
-          <option value="2">Second option</option>
-          <option value="3">Third option</option>
-          <br>
-          <input type="submit" value="Submit" style="font-weight: bold">
-        </select>
+        <form id="chosen-place-form" class="form-element-toolbar">
```
szwbar, line 126: (Not for this PR):
a nicer UX would be to actually populate the options with the restaurants' names. Shouldn't be too hard.

project/src/main/webapp/script.js
```diff
@@ -400,3 +407,29 @@ function registerUserByToken() {
...
+  }
+  let params = [
+    `idToken=${getUserIdToken()}`,
+    `recommendedPlaces=${recommendedPlaces}`,
+    `chosenPlace=${chosenPlaceId}`,
+    `tryAgain=${tryAgain}`
+  ].join('&');
+  fetch('/feedback?' + params, {method: 'POST'});
+  document.getElementById('user-feedback-container').innerText =
+      'Thank you, your Feedback was received!';
```
szwbar: nit: why Feedback with capital F?

project/src/main/webapp/script.js
```diff
@@ -400,3 +407,29 @@ function registerUserByToken() {
...
+  let indexOfPlaceUserChose = getCheckedValueByElementId('chosen-place-form',
+      'Please select the place that you chose.');
+  if (indexOfPlaceUserChose <= recommendedPlaces.length) {
+    storeUserFeedback(recommendedPlaces, recommendedPlaces[indexOfPlaceUserChose].placeId,
+        false /** user tried again */)
+  }
+}
+
+/** Stores user feedback in the database, by using the feedback servlet. */
+function storeUserFeedback(recommendedPlaces, chosenPlaceId, tryAgain) {
```
szwbar: nit: why "store"? Maybe "update" or "post" would be more meaningful here, as you don't actually store anything (the backend does, but it's not up to the JS code to know that).

szwbar: Maybe clear "recommendedPlaces" after you send the feedback?

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+        "place1,place2,place3" /* recommendedPlaces */,
+        "place1" /* chosenPlace */,
+        "false" /* tryAgain */);
+    when(mockUserVerifier.getUserIdByToken(ID_TOKEN)).thenReturn(Optional.of(USER_ID));
+
+    servlet.doPost(REQUEST, RESPONSE);
+
+    // The parameters here must match those that are mocked in the request.
+    // We can't compare instances of UserFeedback, because of the "feedbackTimeInMillis" attribute.
+    verify(mockDataAccessor).updateUserFeedback(argThat(feedback -> {
```
szwbar: FYI: you can always create a custom ArugmentMatcher!
Then you can simplify the syntax to:
`updateUserFeedback(argThat(MatchesUserFeedback(USER_ID, ...)))`

But since you only test this once, it's an overkill here :)

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Test
+  public void doPost_invalidToken_badRequestResponseSent() throws Exception {
+    mockValidRequestParams();
+    when(REQUEST.getParameter("idToken")).thenReturn(null);
+
+    servlet.doPost(REQUEST, RESPONSE);
+
+    verify(mockDataAccessor, never()).updateUserFeedback(any(UserFeedback.class));
```
szwbar, line 101: nice!

szwbar (commented):
Looks really great!

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: I got a linter error about it (shadowing the variable). Is there a better convention to use?

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: Mainly because I wasn't sure how to do it with updating the response. I'll try to figure it out :)

project/src/main/webapp/script.js
karenav: I used "update" and Tal said in the comments here: "I think the name of this function is a little confusing. It stores the feedback right? It doesn't update a feedback already existing. So maybe change to storeUserFeedback."
She has a point... I mean, the function is called in order to start the process of storing the data. 
I think maybe post is a good idea and clearer than the others? WDYT?

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
karenav: Right, I used what Tal did as an example, and I saw that's how she did it. I figured that since it's used once, it would we simpler this way

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: thanks! done

project/src/main/webapp/script.js
karenav: by mistake, thank you! done

project/src/main/webapp/script.js

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
karenav: thank you!

project/src/main/webapp/index.html
karenav: Good idea, I added an issue

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: I did it the way we discussed :)

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: done! :)

project/src/main/webapp/script.js
karenav: done, used post :)

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
```diff
@@ -28,21 +28,13 @@
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 
-/** A servlet that registers a user, according to a given token. */
+/** A servlet that is responsible on storing users' feedback. */
```
szwbar: nit: responsible for

project/src/main/java/com/google/sps/servlets/TokenValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    Optional<String> optionalUserId = userVerifier.getUserIdByToken(userIdToken);
+    if (!optionalUserId.isPresent()) {
+      response.sendError(HttpServletResponse.SC_NOT_FOUND,
+          "Couldn't verify user token (" + validationPurpose + ")");
+      return Optional.empty();
+    }
+    return optionalUserId;
+  }
+
+  private TokenValidator() { }
```
szwbar, line 61: Great!

project/src/main/java/com/google/sps/servlets/TokenValidator.java
```diff
@@ -0,0 +1,58 @@
...
+public final class TokenValidator {
+
+  /**
+   * @param request the HttpServletRequest that we want to check
+   * @param response the HttpServletResponse that is updated in case of invalidity
+   * @param userVerifier the verifier that is used for verifing the user token
+   * @param validationPurpose the reason for turning the token to user ID
+   * @throws IllegalArgumentException if the given request has missing inputs (no token)
+   * @throws IOException if updating the response encounters a problem
+   * @return An Optional with the valid ID if the token is valid, an empty Optional else
```
szwbar: nit: else -> otherwise

project/src/main/java/com/google/sps/servlets/TokenValidator.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.data.UserVerifier;
+import java.util.Optional;
+
+/**
+ * A utility class for validating user tokens.
+ */
+public final class TokenValidator {
+
+  /**
```
szwbar, line 30: Please add a sentence explaining what this function does. Make sure to state that this function modifies the response in case of invalid token ID.

project/src/test/java/com/google/sps/servlets/FeedbackServletTest.java
```diff
@@ -128,4 +145,15 @@ private void mockValidRequestParams(String idToken, String recommendedPlaces, St
     when(REQUEST.getParameter("chosenPlace")).thenReturn(chosenPlace);
     when(REQUEST.getParameter("tryAgain")).thenReturn(tryAgain);
   }
+
+  private static ArgumentMatcher<UserFeedback> matchesUserFeedback(final String userId,
+      final ImmutableList<String> recommendedPlaces, final Optional<String> chosenPlace,
+      final boolean triedAgain) {
+    return feedback ->
```
szwbar, line 152: Cool that you made this a lambda! 👍

szwbar (approved):
Looks good! Maybe add tests for the Token validator as well.

project/src/main/java/com/google/sps/servlets/FeedbackServlet.java
karenav: right, done

project/src/main/java/com/google/sps/servlets/TokenValidator.java
karenav: thanks, done

project/src/main/java/com/google/sps/servlets/TokenValidator.java
karenav: Good point, done

